PR #72, opened 2020-03-16
refactor: Externalize Aries JS Startup opts

rolsonquadras:
- Support to provide Aries JS Opts through a endpoint hosted in server which serves the VueJS app

Signed-off-by: Rolson Quadras <rolson.quadras@securekey.com>


sudeshrshetty:
Just curious, why can't we have this external server in `edge-sandbox`?

rolsonquadras:
> Just curious, why can't we have this external server in `edge-sandbox`?

The user-agent is bundled with http-server. ie, one image for the consumers. They can just set the docker env properties. No extra things required for the consumers to use the user-agent ie, no need to configure external servers.

cmd/http-server/startcmd/start.go
```diff
@@ -95,43 +156,65 @@ func createStartCmd(srv server) *cobra.Command {
 		Short: "Start http server",
 		Long:  "Start http server",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			hostURL, err := getUserSetVar(cmd, hostURLFlagName, hostURLEnvKey, false)
-			if err != nil {
-				return err
+			hostURL, hostURLErr := getUserSetVar(cmd, hostURLFlagName, hostURLEnvKey, false)
```
troyronda: it would be nice to use `err`

cmd/user-agent/src/main.js
```diff
@@ -32,17 +35,44 @@ async function loadAriesOnce() {
...
     return {
-        assetsPath: "/aries-framework-go/assets",
-        "agent-default-label": "dem-js-agent",
-        "http-resolver-url": [],
-        "auto-accept": true,
-        "outbound-transport": ["ws", "http"],
-        "transport-return-route": "all",
-        "log-level": "debug",
-        "db-namespace": "agent"
+        assetsPath: defaultAriesStartupOpts['assetsPath'],
```
troyronda: so the defaults are the DEV options?

cmd/user-agent/src/main.js
```diff
@@ -32,17 +35,44 @@ async function loadAriesOnce() {
...
+}
+
+async function ariesStartupOpts() {
+    let startupOpts = {}
+
+    if (process.env.NODE_ENV === "production") {
+        const axios = require('axios').default;
+
+        // call service to get the opts
+        await axios.get(window.location.origin + '/aries/jsopts')
```
troyronda: TODO: probably could improve the path...

cmd/http-server/startcmd/start.go
rolsonquadras: I updated existing code to increase the patch coverage above 85% (since we have couple of files in the project).

cmd/user-agent/src/main.js
rolsonquadras: yeah, plus which are not configurable externally - since path is part of Vue project.

cmd/user-agent/src/main.js
rolsonquadras: sure. any suggestions ?

PR #149, opened 2020-05-04
fix: remove gunzip for wasm

fqutishat:
Signed-off-by: Firas Qutishat <firas.qutishat@securekey.com>

### DIFF
--- a/cmd/user-agent/scripts/setup_assets.sh
+++ b/cmd/user-agent/scripts/setup_assets.sh
@@ -9,11 +9,9 @@ rm -rf public/aries-framework-go/
 mkdir -p public/aries-framework-go/assets
 npm install
 cp -Rp node_modules/@hyperledger/aries-framework-go/dist/assets/* public/aries-framework-go/assets
-gunzip public/aries-framework-go/assets/aries-js-worker.wasm.gz
 
 
 rm -rf public/trustbloc-agent/
 mkdir -p public/trustbloc-agent/assets
 npm install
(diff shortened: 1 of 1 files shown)


cmd/user-agent/scripts/setup_assets.sh
```diff
@@ -9,11 +9,9 @@ rm -rf public/aries-framework-go/
 mkdir -p public/aries-framework-go/assets
 npm install
 cp -Rp node_modules/@hyperledger/aries-framework-go/dist/assets/* public/aries-framework-go/assets
-gunzip public/aries-framework-go/assets/aries-js-worker.wasm.gz
```
troyronda, line 12: technically the module has a fallback to the original file (aries-js-worker.wasm)
https://github.com/lpar/gzipped/blob/master/fileserver.go#L129

I am guessing that if a browser doesn't send accept-encoding:zip, a 404 will now result.

fqutishat: I'm not sure

troyronda: to support those clients:

```
cp public/aries-framework-go/assets/aries-js-worker.wasm.gz public/aries-framework-go/assets/aries-js-worker.wasm.gz.bak
gunzip public/aries-framework-go/assets/aries-js-worker.wasm.gz
mv public/aries-framework-go/assets/aries-js-worker.wasm.gz.bak public/aries-framework-go/assets/aries-js-worker.wasm.gz
```

PR #163, opened 2020-05-08
feat: DidExchange allows establishing did-connection through the router

soluchok:
This PR includes:
*  Register/Unregister the router (form on UI)
* allows accepting the invitation (button on UI)
* allows accepting the exchange request (button on UI)

NOTE: need to deploy a router

Screenshot_20200508_230713
Screenshot_20200509_132838
Screenshot_20200509_132936

### DIFF
--- a/cmd/user-agent/src/pages/Connections.vue
+++ b/cmd/user-agent/src/pages/Connections.vue
@@ -24,6 +24,45 @@ SPDX-License-Identifier: Apache-2.0
                 <div class="md-layout-item">
                     <md-card class="md-card-plain">
                         <md-card-header data-background-color="green">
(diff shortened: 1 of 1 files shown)


cmd/user-agent/.env.production
```diff
@@ -0,0 +1 @@
+VUE_APP_ROUTER_HOST=http://localhost:10093
```
fqutishat, line 1: why we need .env.production

soluchok: we need a router URL in our vue.js code, to access this variable we need a config file.

soluchok: agree, it doesn't help

PR #197, opened 2020-06-08
feat: didconnection credential type with proof

### LINKED ISSUE #194: WebWallet: DID connection VC should have proof

sudeshrshetty:
- Closes #194

Signed-off-by: sudesh.shetty <sudesh.shetty@securekey.com>

### DIFF
--- a/cmd/user-agent/src/pages/chapi/wallet/get/didConn.js
+++ b/cmd/user-agent/src/pages/chapi/wallet/get/didConn.js
@@ -66,31 +66,55 @@ export class DIDConn {
 
 
     async _didConnResponse(connection) {
-        // TODO add proof to DIDConnection [Issue #194]
+
+        let credential = {
+            "@context": [
+                "https://www.w3.org/2018/credentials/v1",
+                "https://trustbloc.github.io/context/vc/examples-ext-v1.jsonld"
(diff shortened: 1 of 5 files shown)


cmd/user-agent/src/pages/chapi/wallet/get/didConn.js
```diff
@@ -66,31 +66,55 @@ export class DIDConn {
...
+            ],
+            issuer: this.walletUser.did,
+            type: ["VerifiableCredential", "DIDConnection"],
+            credentialSubject: {
+                id: connection.ConnectionID,
+                threadID: connection.ThreadID,
+                inviteeDID: connection.MyDID,
+                inviterDID: connection.TheirDID,
+                inviterLabel: connection.TheirLabel,
+                connectionState: connection.State,
```
rolsonquadras, line 83: I think connectionState and connectionID will not be useful for the other agent as this is specific to the wallet.

sudeshrshetty: connection state we are leaving for now, to track "requested" state when auto accept if off.
credential subject needs a unique ID, reused connection ID for now. Can be filtered later

PR #226, opened 2020-07-10
feat: Use log level command line parameter for TrustBloc agent code

DRK3:

### DIFF
--- a/cmd/user-agent/src/main.js
+++ b/cmd/user-agent/src/main.js
@@ -80,10 +80,11 @@ async function ariesStartupOpts() {
     }
 }
 
-let defaulTrustBlocStartupOpts = {
+let defaultTrustBlocStartupOpts = {
     assetsPath: '/trustbloc-agent/assets',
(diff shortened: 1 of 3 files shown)


cmd/user-agent/src/main.js
```diff
@@ -80,10 +80,11 @@ async function ariesStartupOpts() {
     }
 }
 
-let defaulTrustBlocStartupOpts = {
+let defaultTrustBlocStartupOpts = {
     assetsPath: '/trustbloc-agent/assets',
     blocDomain: 'testnet.trustbloc.local',
-    walletMediatorURL: ''
+    walletMediatorURL: '',
+    'log-level': 'debug'
```
DRK3, line 87: I noticed that the Aries agent uses a default log level of "debug" as the default... so I used "debug" here as well for consistency. Was the choice of "debug" for the aries agent intentional? Or should I change it to "info" (which means I can just remove this parameter since info is the default anyway)?

@sudeshrshetty @fqutishat @rolsonquadras

PR #264, opened 2020-08-04
feat: Added dev mode

soluchok:
Added `dev mode` support. Now, we have a button to switch between modes. Switching between modes triggers an event. So, we can listen to events and draw stuff according to the chosen mode.

Screens:
Screen Shot 2020-08-04 at 6 08 48 PM Screen Shot 2020-08-04 at 6 08 37 PM


cmd/user-agent/src/pages/Layout/DashboardLayout.vue
```diff
@@ -60,8 +60,15 @@ export default {
...
   },
+  mounted: function () {
+    let _this = this;
+    this.$root.$on('dev_mode', function (val) {
+      _this.dev_mode = val
+    })
+  },
   data() {
     return {
+      dev_mode: false,
```
llorllale, line 71: this should be set in main.js

PR #273, opened 2020-08-07
refactor: Added vuex

soluchok:
This PR includes:
* Aries startup loader
* Added vuex support
* refactored `Connections` and `Relationships` 

### DIFF
--- a/cmd/user-agent/src/main.js
+++ b/cmd/user-agent/src/main.js
@@ -5,6 +5,7 @@ SPDX-License-Identifier: Apache-2.0
 */
 
 import Vue from 'vue';
+import store from './store'
 import App from './App.vue';
 import VueRouter from "vue-router";
 import routes from "./router/index";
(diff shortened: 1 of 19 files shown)


cmd/user-agent/src/main.js
```diff
@@ -82,7 +65,7 @@ async function ariesStartupOpts() {
 let defaultTrustBlocStartupOpts = {
     assetsPath: '/trustbloc-agent/assets',
     blocDomain: 'testnet.trustbloc.local',
-    walletMediatorURL: '',
+    walletMediatorURL: 'http://localhost:10093',
```
ToJen, line 68: It might be possible to load these as environment variables from the outside: https://cli.vuejs.org/guide/mode-and-env.html#modes

soluchok: I am loading it from the outside.I will rollback this change.

PR #281, opened 2020-08-14
feat: credential mediators for CHAPI

sudeshrshetty:
- added credential mediators for CHAPI
- if not provided, then CHAPI will use default authn.io
- refactored credential handler polyfill settings and web worker
- docker image for credential mediator

TODO: in next PR, TLS options will be added to credential-mediator
docket image

Signed-off-by: sudesh.shetty <sudesh.shetty@securekey.com>


sudeshrshetty:
ignore codecov please, couldn't cover oneline which is very hard to cover.

docs/test/test.md
```diff
@@ -22,6 +22,7 @@ You need to modify your hosts file (`/etc/hosts` on \*NIX) to add the following
     127.0.0.1 testnet.trustbloc.local
     127.0.0.1 stakeholder.one
     127.0.0.1 sidetree-mock
+    127.0.0.1 authorization.localhost
```
rolsonquadras, line 25: Should this be `authorization.local` to be consistent with other DNS names we are using (eg, `testnet.trustbloc.local`)?

soluchok: @sudeshrshetty domain should be added here as well
https://github.com/trustbloc/edge-agent/blob/master/scripts/generate_test_keys.sh#L24

sudeshrshetty: @soluchok as I mentioned in TODO, this one will be added in next PR while enabling TLS.

sudeshrshetty: @rolsonquadras using this [domain name](https://github.com/digitalbazaar/authorization.io/blob/57e33cb4b3d251e02ed5b179068ccda51560b26d/lib/config.js#L29) for now, will try to update it next PR

PR #290, opened 2020-08-18
chore: update trustbloc-did-method dependency

Moopli:
Signed-off-by: Filip Burlacu <filip.burlacu@securekey.com>


Makefile
```diff
@@ -94,15 +94,19 @@ generate-test-keys:
 		frapsoft/openssl
 
 .PHONY: user-agent-start
-user-agent-start: clean user-agent-wasm-docker credential-mediator-docker generate-test-config generate-test-keys
+user-agent-start: clean user-agent-wasm-docker credential-mediator-docker did-method-cli generate-test-keys
```
sudeshrshetty, line 97: We don't need `generate-test-config` anymore?
if Yes, then why that target still exists in code base?

scripts/check_js_intergation.sh
```diff
@@ -13,7 +13,12 @@ ROOT=`pwd`
 npm -v
 echo "starting containers..."
 cd $ROOT/test/bdd/fixtures/agent-wasm
-(source .env && docker-compose down && docker-compose up --force-recreate -d)
+
+(source .env && docker-compose down --remove-orphans && docker-compose -f docker-compose.yml up --force-recreate -d)
+
+bash generate_config.sh
```
sudeshrshetty, line 19: Why we need it only for bddtests and not for user-agent-start?

Makefile
Moopli: Should have removed it, fixed.

scripts/check_js_intergation.sh
Moopli: fixed.

sudeshrshetty: not sure what did you fix here

Moopli: I added this to scripts/user_agent_start.sh, so now user-agent-start also splits the docker-compose and generates the configs.

scripts/user_agent_start.sh
```diff
@@ -9,4 +9,9 @@ set -e
 echo "Running $0"
 
 cd test/bdd/fixtures/agent-wasm
-(source .env && docker-compose down && docker-compose up --force-recreate)
\ No newline at end of file
+
+(source .env && docker-compose down --remove-orphans && docker-compose -f docker-compose.yml up --force-recreate -d)
+
+bash generate_config.sh
```
Moopli, line 15: @sudeshrshetty

PR #295, opened 2020-08-19
feat: issuer manifest is matched if the credential schema uri is specified in a `const` filter

llorllale:
This change is really more of a temporary (and cheap) solution to support [this](https://github.com/trustbloc/edge-adapter/pull/252) change in the RP edge-adapter.

I say this is *temporary* because I think a better solution would be to extend the [IssuerManifestCredential](https://github.com/trustbloc/context/blob/master/vc/issuer-manifest-credential-v1.jsonld) such that `contexts` is an array of VCs or VC-like objects against which the wallet and Issuer can evaluate the presentation-exchange criteria. This change would align better with [this existing TODO](https://github.com/trustbloc/edge-agent/blob/5b8834b8fe1ac848080f81b3c27ef46d22e65877/cmd/user-agent/src/pages/chapi/wallet/common/presentationExchange.js#L208).

### DIFF
--- a/cmd/user-agent/src/pages/chapi/wallet/common/presentationExchange.js
+++ b/cmd/user-agent/src/pages/chapi/wallet/common/presentationExchange.js
@@ -210,6 +210,12 @@ function matchManifest(manifest, descriptor) {
 
     let schemas = Array.isArray(descriptor.schema.uri) ? descriptor.schema.uri : [descriptor.schema.uri]
 
+    if (descriptor.constraints && descriptor.constraints.fields) {
+        descriptor.constraints.fields.filter(f => f.filter).filter(f => f.filter.const).forEach(f => {
+            schemas.push(f.filter.const)
+        })
+    }
+
     return manifest.credentialSubject.contexts.some(v => schemas.includes(v))
 }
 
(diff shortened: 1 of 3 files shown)


cmd/user-agent/src/pages/chapi/wallet/common/presentationExchange.js
```diff
@@ -210,6 +210,13 @@ function matchManifest(manifest, descriptor) {
 
     let schemas = Array.isArray(descriptor.schema.uri) ? descriptor.schema.uri : [descriptor.schema.uri]
 
+    // yuck
```
rolsonquadras, line 213: Can we add a TODO here to remove `yuck` ? :)

llorllale: I wanted to [coallesce the null checks](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and tried adding [this](https://www.npmjs.com/package/@babel/plugin-proposal-optional-chaining) dependency but couldn't for unclear reasons.... I think the TODO a couple of lines above will likely refactor the whole function anyway.

llorllale: Removed just to avoid the extra noise

fqutishat (approved):

Issue #267, opened 2020-08-04
SDS integration: Encrypt DIDs, credentials and presentations before storing


DRK3:
Once edge-agent has encryption keys available, actually encrypt the documents being stored in SDS.

rolsonquadras:
This has been implemented

PR #297, opened 2020-08-19
fix: Added default SDS URL and agent username TrustBloc agent opts

### LINKED ISSUE #296: Add default SDS and agent username TrustBloc startup opts
The SDS default startup opt in particular will allow user-agent Javascript code to be run with "npm run serve" without SDS-related runtime errors. The agent username is needed for SDS, and shouldn't be blank either.

DRK3:
closes #296

- user-agent JavaScript code can now be run with "npm run serve" without SDS-related runtime errors.

- Added blank SDS URL check to sdscomm. This prevents an unintuitive error message from being produced.

- Added blank agent username check to sdscomm.

### DIFF
--- a/cmd/user-agent/src/main.js
+++ b/cmd/user-agent/src/main.js
@@ -69,6 +69,8 @@ let defaultTrustBlocStartupOpts = {
     'log-level': 'debug',
     walletMediatorURL: '',
     credentialMediatorURL: '',
+    agentUsername: 'user-agent',
+    sdsServerURL: 'https://localhost:8072/encrypted-data-vaults'
 }
 
 async function trustblocStartupOpts() {
@@ -93,8 +95,8 @@ async function trustblocStartupOpts() {
         walletMediatorURL: ('walletMediatorURL' in startupOpts) ? startupOpts['walletMediatorURL'] : defaultTrustBlocStartupOpts['walletMediatorURL'],
         credentialMediatorURL: credentialMediator(('credentialMediatorURL' in startupOpts) ? startupOpts['credentialMediatorURL'] : defaultTrustBlocStartupOpts['credentialMediatorURL']),
         'log-level': ('log-level' in startupOpts) ? startupOpts['log-level'] : defaultTrustBlocStartupOpts['log-level'],
-        agentUsername: startupOpts['agentUsername'],
-        sdsServerUrl: startupOpts['sdsServerURL']
+        agentUsername: ('agentUsername' in startupOpts) ? startupOpts['agentUsername'] : defaultTrustBlocStartupOpts['agentUsername'],
+        sdsServerURL: ('sdsServerURL' in startupOpts) ? startupOpts['sdsServerURL'] : defaultTrustBlocStartupOpts['sdsServerURL']
     }
 }
 
--- a/pkg/controller/command/didclient/command.go
+++ b/pkg/controller/command/didclient/command.go
@@ -43,6 +43,7 @@ const (
 	failDecodeDIDDocDataErrMsg = "failed to decode DID data"
 	failCreateDIDVaultErrMsg   = "failed to create DID vault"
 	failStoreDIDDocErrMsg      = "failed to store DID document"
+	failCreateSDSCommErrMsg    = "failure while preparing SDS communication"
 )
 
 type didBlocClient interface {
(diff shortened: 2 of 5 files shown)


cmd/user-agent/src/main.js
```diff
@@ -69,6 +69,8 @@ let defaultTrustBlocStartupOpts = {
     'log-level': 'debug',
     walletMediatorURL: '',
     credentialMediatorURL: '',
+    agentUsername: 'user-agent',
+    sdsServerURL: 'https://localhost:8072/encrypted-data-vaults'
```
sudeshrshetty, line 73: don't hardcode SDS URL in default opts, leave it as empty.

pkg/controller/command/didclient/command.go
```diff
@@ -149,9 +150,13 @@ func (c *Command) SaveDID(_ io.Writer, req io.Reader) command.Error {
 }
 
 func (c *Command) saveDID(didDataToStore *sdscomm.DIDDocData) command.Error {
-	sdsComm := sdscomm.New(c.sdsServerURL, c.agentUsername)
+	sdsComm, err := sdscomm.New(c.sdsServerURL, c.agentUsername)
```
sudeshrshetty, line 153: why are you creating new sdsComm instance everytime? looks like serverURL & agentUsername are always same here?

pkg/controller/command/sdscomm/sdscomm.go
```diff
@@ -34,12 +38,21 @@ type DIDDocData struct {
 	Name     string          `json:"name,omitempty"`
 }
 
-func New(sdsServerURL, agentUsername string) *SDSComm {
+func New(sdsServerURL, agentUsername string) (*SDSComm, error) {
+	logger.Infof("!!!DEREK about to check if the provided SDS URL is blank.")
```
sudeshrshetty, line 42: `DEREK` ???

cmd/user-agent/src/main.js
llorllale: @sudeshrshetty how is this different than the hard-coding in `defaultAriesStartupOpts` or the existing keys in `defaultTrustBlocStartupOpts`?

Part of this change also makes the SDS URL mandatory otherwise it will fail. Adding a toggle to choose local or remote storage is still a ways off into the future.

pkg/controller/command/sdscomm/sdscomm.go
DRK3: Oops... fixed :)

pkg/controller/command/didclient/command.go
DRK3: Every time the go code executes it creates a new Command object. Is there a way I can avoid this?

cmd/user-agent/src/main.js
sudeshrshetty: Default ops won't work. Even walletMediatorURL is mandatory, we didn't set it. 

If we want edge-agent to work with default ops, then we need to initialize all required opts properly.

pkg/controller/command/sdscomm/sdscomm.go
DRK3: (just removed the log line - didn't mean to commit that...)

cmd/user-agent/src/main.js
DRK3: @sudeshrshetty Note that at least some of the edge-agent functionality works without those other opts (but with the SDS URL). I tested registering wallet and DID creation and they seemed to work ok.

I can leave it as blank URL if you'd really prefer that - I just figured that it's a little more convenient so that you can do UI/JS work quickly. At least now with this change you'll get a more obvious error message.

llorllale: Yeah I am trying to understand what @sudeshrshetty is saying.... I see that `trustblocStartupOpts()` pulls the config from the backend... yet still uses `defaultTrustBlocStartupOpts` if `walletMediatorURL` is not present...

So `defaultTrustBlocstartupOpts` is the *fallback* and should definitely have *something* there. Config optional in the backend, mandatory in the frontend, from the looks of it.

sudeshrshetty: I agree @llorllale , then we should initialize it with default options for all mandatory options.

pkg/controller/command/didclient/command.go
sudeshrshetty: Are you sure it creates new command object for each call??
Still, you should move this to constructor(New()) function

cmd/user-agent/src/main.js
DRK3: @sudeshrshetty Should I create a follow-up story for that and leave the `https://localhost:8072/encrypted-data-vaults` arg as is?

pkg/controller/command/didclient/command.go
DRK3: @sudeshrshetty Yeah, I actually did some testing since I was curious about exactly this... in fact, controller.GetCommandHandlers inside the addTrustBlocAgentHandlers function seemed to be called new every time I clicked...

Since this change seems unrelated to the current PR and will cause merge conflicts with the other PR, I'll fix this in a separate PR after my current two PRs are merged in. #267

DRK3: Clarification: I tested this with the Create DID button on the DID management page

cmd/user-agent/src/main.js
sudeshrshetty: @DRK3 not needed for now

PR #298, opened 2020-08-20
chore: Revert "chore: update trustbloc-did-method dependency"

Moopli:
Reverts trustbloc/edge-agent#290


Moopli:
Going to reopen with fixed DCO and semantic commit

PR #316, opened 2020-08-27
feat: Make SDS server URL and agent label optional for TrustBloc agent

DRK3:

### DIFF
--- a/cmd/user-agent/src/main.js
+++ b/cmd/user-agent/src/main.js
@@ -69,8 +69,8 @@ let defaultTrustBlocStartupOpts = {
     'log-level': 'debug',
     walletMediatorURL: 'https://localhost:10063',
     credentialMediatorURL: '',
-    agentUsername: 'user-agent',
-    sdsServerURL: 'https://localhost:8072/encrypted-data-vaults'
+    agentUsername: '',
+    sdsServerURL: ''
 }
 
 async function trustblocStartupOpts() {
--- a/cmd/user-agent/src/pages/chapi/wallet/didmgmt/didManager.js
+++ b/cmd/user-agent/src/pages/chapi/wallet/didmgmt/didManager.js
@@ -97,14 +97,18 @@ export class DIDManager extends KeyValueStore {
             }
         )
 
-        const t = await new this.trustblocAgent.Framework(this.trustblocStartupOpts)
+        if (this.trustblocStartupOpts.sdsServerURL) {
+            const t = await new this.trustblocAgent.Framework(this.trustblocStartupOpts)
 
-        // Save DID to persistent storage
-        await t.didclient.saveDID({
-            name: name,
-            signType: signType,
-            did: did
-        })
+            // Save DID to persistent storage
+            await t.didclient.saveDID({
+                name: name,
+                signType: signType,
+                did: did
+            })
+        } else {
+            console.log("Skipping DID storage to SDS since no SDS server URL was configured.")
+        }
     }
 
     async getAllDIDMetadata() {
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -7,8 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package controller
 
 import (
-	"fmt"
-
 	"github.com/trustbloc/edge-core/pkg/log"
 
 	"github.com/trustbloc/edge-agent/pkg/controller/command"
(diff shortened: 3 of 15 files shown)


pkg/controller/controller.go
```diff
@@ -63,10 +57,7 @@ func GetCommandHandlers(opts ...Opt) ([]command.Handler, error) {
 		opt(cmdOpts)
 	}
 
-	sdsComm, err := sdscomm.New(cmdOpts.sdsServerURL, cmdOpts.agentUsername)
-	if err != nil {
-		return nil, fmt.Errorf(failCreateSDSCommErrMsg, err)
-	}
+	sdsComm := sdscomm.New(cmdOpts.sdsServerURL, cmdOpts.agentUsername)
```
DRK3, line 60: It seems that every time a WASM command gets executed, this GetCommandHandlers function gets executed as well. So in order to prevent other WASM commands from getting an error due to a blank sds server url, I moved the check into the specific command functions.

pkg/controller/command/presentationclient/command.go
```diff
@@ -52,6 +56,11 @@ func (c *Command) GetHandlers() []command.Handler {
 }
 
 func (c *Command) SavePresentation(_ io.Writer, req io.Reader) command.Error {
+	if c.sdsComm.SDSServerURL == "" {
+		logutil.LogError(logger, commandName, savePresentationCommandMethod, failToStorePresentationNoSDSURLSet)
+		return command.NewValidationError(InvalidRequestErrorCode, errFailToStorePresentationNoSDSURLSet)
+	}
```
DRK3, line 62: This is just an extra layer of safety in case this method gets called somehow (perhaps by another dev in the future). Without this check, the error message is super confusing.

pkg/controller/command/didclient/command.go
```diff
@@ -126,6 +131,11 @@ func (c *Command) CreateDID(rw io.Writer, req io.Reader) command.Error {
 }
 
 func (c *Command) SaveDID(_ io.Writer, req io.Reader) command.Error {
+	if c.sdsComm.SDSServerURL == "" {
+		logutil.LogError(logger, commandName, saveDIDCommandMethod, failToStoreDIDNoSDSURLSet)
+		return command.NewValidationError(InvalidRequestErrorCode, errFailToStoreDIDNoSDSURLSet)
+	}
```
DRK3, line 137: This is just an extra layer of safety in case this method gets called somehow (perhaps by another dev in the future). Without this check, the error message is super confusing.

pkg/controller/command/credentialclient/command.go
```diff
@@ -52,6 +56,11 @@ func (c *Command) GetHandlers() []command.Handler {
 }
 
 func (c *Command) SaveCredential(_ io.Writer, req io.Reader) command.Error {
+	if c.sdsComm.SDSServerURL == "" {
+		logutil.LogError(logger, commandName, saveCredentialCommandMethod, failToStoreCredentialNoSDSURLSet)
+		return command.NewValidationError(InvalidRequestErrorCode, errFailToStoreCredentialNoSDSURLSet)
+	}
```
DRK3, line 62: This is just an extra layer of safety in case this method gets called somehow (perhaps by another dev in the future). Without this check, the error message is super confusing.

fqutishat: We are not suppose to add check in every method

DRK3: Moved the check to a common spot in sdsComm.

cmd/user-agent/src/pages/chapi/wallet/didmgmt/didManager.js
```diff
@@ -97,14 +97,18 @@ export class DIDManager extends KeyValueStore {
...
+            const t = await new this.trustblocAgent.Framework(this.trustblocStartupOpts)
 
-        // Save DID to persistent storage
-        await t.didclient.saveDID({
-            name: name,
-            signType: signType,
-            did: did
-        })
+            // Save DID to persistent storage
+            await t.didclient.saveDID({
```
rolsonquadras, line 104: Are we not even storing in indexeddb if `sds` is not configured ? Or is there another save call to indexeddb ?

cmd/user-agent/src/main.js
```diff
@@ -69,8 +69,8 @@ let defaultTrustBlocStartupOpts = {
     'log-level': 'debug',
     walletMediatorURL: 'https://localhost:10063',
     credentialMediatorURL: '',
-    agentUsername: 'user-agent',
-    sdsServerURL: 'https://localhost:8072/encrypted-data-vaults'
+    agentUsername: '',
+    sdsServerURL: ''
```
DRK3, line 73: Now that SDS is optional, there will be no errors if this is blank. Also note that we had decided that the default setting should be off while the edge-agent<->SDS flow is being developed.

cmd/user-agent/src/pages/chapi/wallet/didmgmt/didManager.js
DRK3: Yeah, there's another save call to local browser storage just above this - it's cut off by the diff.

PR #326, opened 2020-09-02
feat: Use username from login for SDS vault creation

### LINKED ISSUE #266: SDS integration: Get username from user registration instead of cmd line arg
In order to get some SDS integration working early on, I used the agent label as user identifier for creating vaults. This username should come from the actual user registration when that part is done/ ready to use.

DRK3:
closes #266

### DIFF
--- a/cmd/user-agent/src/pages/chapi/wallet/register/register.js
+++ b/cmd/user-agent/src/pages/chapi/wallet/register/register.js
@@ -36,6 +36,11 @@ export class RegisterWallet extends WalletManager {
         // create DID
         let did = await this.didManager.createDID(keyType, signType)
 
+        // Make sure at least the username is stored so that the DID can be saved to SDS
+        // This will be overridden by another call at the end of the function that contains the full wallet metadata
+        // TODO: Refactor to avoid calling this twice in the same function. #332
+        await this.storeWalletMetadata(user, {})
+
         // save DID
         await this.didManager.saveDID(`${user}_${uuid()}`, signType, did)
 
--- a/pkg/controller/command/sdscomm/models.go
+++ b/pkg/controller/command/sdscomm/models.go
@@ -8,18 +8,21 @@ package sdscomm
 
 import "encoding/json"
 
-type DIDDocData struct {
+type SaveDIDDocToSDSRequest struct {
 	Name     string          `json:"name,omitempty"`
 	SignType string          `json:"signType,omitempty"`
 	DID      json.RawMessage `json:"did,omitempty"`
+	UserID   string          `json:"userID,omitempty"`
 }
 
-type CredentialData struct {
+type SaveCredentialToSDSRequest struct {
 	Name       string          `json:"name,omitempty"`
 	Credential json.RawMessage `json:"credential,omitempty"`
+	UserID     string          `json:"userID,omitempty"`
 }
 
-type PresentationData struct {
+type SavePresentationToSDSRequest struct {
 	Name         string          `json:"name,omitempty"`
 	Presentation json.RawMessage `json:"presentation,omitempty"`
+	UserID       string          `json:"userID,omitempty"`
 }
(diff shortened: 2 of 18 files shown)


DRK3:
> The vault id cannot be a username in the real system. Needs to be a random globally unique identifier. Like a GUID.
> 
> I also think EDV should be generating the vault names during vault creation. It shouldn’t allow the client to pick the name.

This is only temporary. This will be fixed later: https://github.com/trustbloc/edge-agent/issues/265

cmd/user-agent/src/pages/chapi/Login.vue
```diff
@@ -60,11 +62,22 @@ SPDX-License-Identifier: Apache-2.0
             if (wuser) {
                 this.registered = true
                 this.statusMsg = `Wallet already registered for ${wuser.id}`
+            } else {
```
sudeshrshetty, line 65: what's the motive behind this logic?

sudeshrshetty (commented):
Can you elaborate these changes?

cmd/user-agent/src/pages/chapi/Login.vue
sudeshrshetty: all these logics in login vue is gone now, and in future this gonna be replaced by hub-auth page redirect.
Please rebase.

sudeshrshetty (changes requested):

cmd/user-agent/src/pages/chapi/Login.vue
DRK3: Rebased. Originally I had that logic in order to pre-populate the username field based on the previously used username, and I also had it generating a random username for new users. These were just nice-to-haves while it was in development, and would make username collisions harder... but seeing as how it's going to be moved soon it's probably not that important at the moment. I removed that extra logic.

cmd/user-agent/src/pages/chapi/wallet/register/register.js
```diff
@@ -36,6 +36,10 @@ export class RegisterWallet extends WalletManager {
         // create DID
         let did = await this.didManager.createDID(keyType, signType)
 
+        // Make sure at least the username is stored so that the DID can be saved to SDS
+        // This will be overridden by another call at the end of the function that contains the full wallet metadata
+        await this.storeWalletMetadata(user, {})
```
sudeshrshetty, line 41: saving jsindex db twice for saving same record in single function is a performance bottleneck, add a followup TODO to avoid this. May be a better flow would save a db call here.

pkg/controller/command/sdscomm/models.go
```diff
@@ -8,18 +8,21 @@ package sdscomm
...
 
-type DIDDocData struct {
-	Name     string          `json:"name,omitempty"`
-	SignType string          `json:"signType,omitempty"`
-	DID      json.RawMessage `json:"did,omitempty"`
+type SaveDIDDocToSDSRequest struct {
+	Name          string          `json:"name,omitempty"`
+	SignType      string          `json:"signType,omitempty"`
+	DID           json.RawMessage `json:"did,omitempty"`
+	AgentUsername string          `json:"agentUsername,omitempty"`
```
sudeshrshetty, line 15: may be a followup to have better name, 'userid, username, user etc', term 'agent' is misleading. Same applies to other requests in this file.

cmd/user-agent/src/pages/chapi/wallet/register/register.js
sudeshrshetty, line 41: saving jsindex db twice for saving same record in single function is a performance bottleneck, add a followup TODO to avoid this. May be a better flow would save a db call here.

pkg/controller/command/sdscomm/models.go
DRK3: Updated to "userID".

cmd/user-agent/src/pages/chapi/wallet/register/register.js
DRK3: Added a follow-up TODO. Perhaps when we add the proper login flow there will be a solution to this... I'll keep an eye out.

sudeshrshetty (approved):

rolsonquadras (approved):

PR #358, opened 2020-09-18
feat: #357 http-server OIDC /login endpoint

### LINKED ISSUE #357: OIDC /login endpoint
part of #353 

Need to implement an endpoint from which the UI can fetch the OIDC authN/Z request.

llorllale:
~~depends on #359~~ 

closes #357 

### DIFF
--- a/pkg/restapi/common/common.go
+++ b/pkg/restapi/common/common.go
@@ -1,0 +1,40 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package common
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/trustbloc/edge-core/pkg/log"
+)
+
+// Handler http handler for each controller API endpoint.
+type Handler interface {
+	Path() string
+	Method() string
+	Handle() http.HandlerFunc
+}
+
+// ErrorResponse to send error message in the response.
+type ErrorResponse struct {
+	Message string `json:"errMessage,omitempty"`
+}
+
+// WriteErrorResponsef write error resp.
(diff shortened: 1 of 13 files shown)


pkg/restapi/oidc/dependencies.go
```diff
@@ -0,0 +1,46 @@
...
+	return v.v.Verify(ctx, token)
+}
+
+func (o *oidcProviderImpl) Endpoint() oauth2.Endpoint {
+	return o.op.Endpoint()
+}
+
+type idToken interface {
+	Claims(interface{}) error
+}
```
Baha-sk, line 46: seems odd to have to create an interface that is only used in returns of this file, I would prefer to use OIDC's interface directly instead but can't seem to find one, is this intentional?

llorllale: The intent is to be able to mock these dependencies in unit tests in order to have good coverage.

Baha-sk (approved):

pkg/restapi/common/common.go
```diff
@@ -0,0 +1,40 @@
...
+// WriteErrorResponsef write error resp.
+func WriteErrorResponsef(rw http.ResponseWriter, logger log.Logger, status int, msg string, args ...interface{}) {
+	rw.WriteHeader(status)
+
+	err := json.NewEncoder(rw).Encode(ErrorResponse{
+		Message: fmt.Sprintf(msg, args...),
+	})
+
+	if err != nil {
+		logger.Errorf("Unable to send error message, %s", err)
```
DRK3, line 38: Just for consistency, consider using a `:` instead of a `,`

DRK3 (approved):
Left minor nitpick (optional)

pkg/restapi/common/common.go
llorllale: done

PR #359, opened 2020-09-18
chore: upgrade to Go 1.15

llorllale:


llorllale:
Low coverage on existing lines of code.

rolsonquadras (approved):

cmd/http-server/startcmd/start.go
```diff
@@ -159,23 +159,26 @@ func VueHandler(publicDir string, opts *ariesJSOpts, trustblocAgentOpts *trustbl
...
-			w.Write(j) // nolint errcheck
 			return
 		}
 
 		// trustbloc agent js opts
 		if urlPath == trustblocStartupOptsPath {
-			j, _ := json.Marshal(trustblocAgentOpts) // nolint errcheck
+			j, _ := json.Marshal(trustblocAgentOpts) // nolint:errcheck // don't know why err is ignored
+
+			w.Write(j) // nolint:errcheck,gosec // don't know why err is ignored
```
DRK3, line 173: Rather than adding a "don't know..." type of comment, it would be better to create a follow-up issue to look at it and then link it here.

DRK3 (changes requested):
See comment

cmd/http-server/startcmd/start.go
llorllale: @DRK3 PR #368 refactored this part and is removing `VueHandler`. We should refactor/do followup there.

DRK3 (approved):

PR #368, opened 2020-09-22
feat: #361 hook OIDC REST handlers into http-server

### LINKED ISSUE #361: hook OIDC REST operation handlers into server
#357 is adds a new `/oidc/login` endpoint.

We need to add this endpoint's handler to the http server.

llorllale:
~~depends on #358~~

closes #361 

**Breaking change**

The wallet is now hosted under `/wallet/`. Users must browse to `https://<host>/wallet/` in order to reach the wallet.

Users must clear browser storage, login again and re-register their wallet.

----

**Testing**

This change was tested against the sandbox with the [New Bank Account Use Case](https://github.com/trustbloc/edge-sandbox/blob/master/docs/demo/new-bank-account-usecase.md) flow. 

----

**Summary**

The wallet's assets are now served under the `/wallet/` path. The wasm startup options are served under `/walletconfig/aries` and `/walletconfig/trustbloc`. The OIDC handlers are served under `/oidc`.

Many links had to be fixed in the UI layer. A best effort was made to centralize base path configuration using `__webpack_public_path__`. The sidebar links remain to be properly configured this way.

----

**Changes**

* UI assets served on `/wallet/` instead of `/`
* aries & trustbloc agent opts served on `/walletconfig/aries` and `/walletconfig/trustbloc` respectively
* OIDC operations served under `/oidc/`
* Vue.js: best effort done to configure base path in all locations using __webpack_public_path__ alone. (pending sidebar links)
* OIDC startup flags added
* CA Certs startup flag added because OIDC providers in sandbox serve HTTPS
    
----


cmd/http-server/startcmd/start.go
```diff
@@ -404,12 +478,91 @@ func startHTTPServer(parameters *httpServerParameters) error {
...
+		http.ServeFile(w, r, path.Join(publicDir, indexHTLMPath))
+	})
+}
+
+func addUIConfigHandlers(router *mux.Router, config *httpServerParameters) {
+	router.HandleFunc("/aries", func(w http.ResponseWriter, r *http.Request) {
+		logger.Infof("handling config request: %s", r.URL.String())
+		bits, _ := json.Marshal(config.opts)
+
+		w.Write(bits)
```
llorllale, line 535: we should not ignore these errors and we should test these lines

DRK3 (approved):
Looks like you need to squash your commits

soluchok (approved):

aholovko (approved):

sudeshrshetty (commented):
Can we have a follow up to rename `/ui` to something like `/wallet`?
it looks odd

cmd/user-agent/src/pages/chapi/Worker.vue
```diff
@@ -26,11 +26,11 @@ SPDX-License-Identifier: Apache-2.0
                 mediatorOrigin : opts.credentialMediatorURL,
                 async get(event) {
                     console.log('Received get() event:', event.event);
-                    return {type: 'redirect', url: '/GetFromWallet'};
+                    return {type: 'redirect', url: `${__webpack_public_path__}/GetFromWallet`};
```
sudeshrshetty, line 29: Just curious: why public path in this URL?

cmd/user-agent/src/router/index.js
```diff
@@ -79,7 +79,24 @@ const routes = [
...
                 meta: {requiresAuth: true}
-            }
+            },
+            {
+                path: "worker",
+                name: "chapi-worker",
+                component: WalletWorker
+            },
+            {
+                path: "StoreInWallet",
```
sudeshrshetty, line 89: These pages shouldn't be part of router children, 
We intentionally kept them out of 'router children' to hide left navigation, top navigation.

sudeshrshetty (changes requested):

llorllale:
@sudeshrshetty 

> Can we have a follow up to rename `/ui` to something like `/wallet`?
> it looks odd

Changed to `/wallet/`

cmd/user-agent/src/router/index.js
llorllale: This question is related to your other question here: https://github.com/trustbloc/edge-agent/pull/368#discussion_r494446766.

These pages need to be served by the backend in order to be reachable by the `authn.io` credential mediator domain.

Here's some logging output from the backend while storing a presentation using `/wallet/webwallet`:

image

llorllale: I don't really understand your comments about "hiding left navigation, top navigation"

llorllale: @sudeshrshetty fixed, here is how the CHAPI layout looks:

image

image

llorllale: @sudeshrshetty disabled the footer on the CHAPI windows as per your request (reused the `hideFooter` route metadata)

sudeshrshetty (approved):

PR #373, opened 2020-09-24
refactor: replace aries-js-worker with in-place agent-js-worker

### LINKED ISSUE #372: Implement support for a custom aries-framework-go wasm
Currently `edge-agent` imports functionality of `aries-js-worker` from the framework directly - and because of that, the only way how we can customize an instance of the framework is thru input parameters.

To be more robust in adjusting the framework to our particular needs (e.g. specifying trustbloc-specific implementations of kms/crypto) a custom "build" of the aries-framework-go wasm should be used instead (so we get access to the integration root, i.e. `aries.New(options...)`).

The goal of the current task is to replace imported `@trustbloc-cicd/aries-framework-go` in `user-agent` with the in-place `agent-js-worker` that replicates the same functionality as `aries-js-worker` from the framework. The next task will be to merge both `agent-js-worker` (result of the current task) and existing `trustbloc-agent-js-worker` into a single package so all "js worker" stuff resides in one place.

aholovko:
Added custom aries-framework-go wasm (`agent-js-worker`) and updated `user-agent` to use it instead of directly importing `@trustbloc-cicd/aries-framework-go`. This is the first step toward configurable aries-framework-go wasm instance in `edge-agent` (the next one is to merge `agent-js-worker` and `trustbloc-agent-js-worker` into a single package).

This is quite a big PR, but most of the code is boilerplate.

Closes #372

Signed-off-by: Andriy Holovko <andriy.holovko@gmail.com>


rolsonquadras (approved):

llorllale (approved):

cmd/agent-js-worker/src/agent.js
```diff
@@ -0,0 +1,1089 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
```
sudeshrshetty, line 6: is this file same as aries.js from aries-framework-go, then add a reference to original file.

cmd/agent-js-worker/src/agent.js
```diff
@@ -0,0 +1,1089 @@
...
+// base path to load assets from at runtime
+const __publicPath = _ => {
+    if (inNode) {
+        // TODO determine module_path at runtime
+        return process.cwd() + "/node_modules/@trustbloc/agent-js-worker/"
+    } else if (inBrowser) {
+        return "/agent-js-worker/"
+    } else {
+        // TODO #1127 - throw error or use default?
```
sudeshrshetty, line 26: 1217 is from aries project,
better a add comment at the top saying this file is copied from aries, users for this js worker should know.

aholovko: ok, that stuff definitely needs additional cleanups - will do that as part of the next task about merging `agent-js-worker` and `trustbloc-agent-js-worker`

cmd/agent-js-worker/src/agent.js
aholovko: it will be extended with functionality from `trustbloc-agent-js-worker`

sudeshrshetty: then it shouldn't be called 'aries-js-worker' once merged

cmd/agent-js-worker/src/agent.js
sudeshrshetty: then it shouldn't be called 'aries-js-worker' once merged

aholovko: it will not, I plan to merge functionality of `trustbloc-agent-js-worker` into `agent-js-worker` so we will end up with the following structure:
```
|-agent-js-worker
|-http-server
|-user-agent 
```

cmd/agent-js-worker/src/agent.js
aholovko: it's **agent**-js-worker not **aries**-js-worker

PR #387, opened 2020-10-01
feat: #379 session cookies

### LINKED ISSUE #379: Session cookies
part of #353 

Use cookies to authenticate the user's session.

This looks good: https://github.com/gorilla/sessions

Good reference: https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#session-management-implementation

llorllale:
closes #379 

This patch adds initial support for user session cookies. The OAuth2 'state' parameter is stored in an encrypted and authenticated form in a session cookie that is then validated against the value received in the 'state' query parameter in the callback.

We use the gorilla/sessions module that allows setting an authentication key and an encryption key in order to authenticate and protect the session cookies stored in the browser. These keys are configured as "session cookie keys" and passed in by an admin as variables that point to a file that contains the raw bytes.

### DIFF
--- a/cmd/http-server/startcmd/start.go
+++ b/cmd/http-server/startcmd/start.go
@@ -11,8 +11,10 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -169,6 +171,19 @@ const (
 	oidcCallbackURLEnvKey = "HTTP_SERVER_OIDC_CALLBACK"
 )
 
+// Keys.
+const (
+	sessionCookieAuthKeyFlagName  = "cookie-auth-key"
+	sessionCookieAuthKeyFlagUsage = "Path to the pem-encoded 32-byte key to use to authenticate session cookies." +
+		" Alternatively, this can be set with the following environment variable: " + sessionCookieAuthKeyEnvKey
+	sessionCookieAuthKeyEnvKey = "HTTP_SERVER_COOKIE_AUTH_KEY"
+
+	sessionCookieEncKeyFlagName  = "cookie-enc-key"
+	sessionCookieEncKeyFlagUsage = "Path to the pem-encoded 32-byte key to use to encrypt session cookies." +
+		" Alternatively, this can be set with the following environment variable: " + sessionCookieEncKeyEnvKey
+	sessionCookieEncKeyEnvKey = "HTTP_SERVER_COOKIE_ENC_KEY"
+)
+
 var logger = log.New("edge-agent/http-server")
 
 type server interface {
@@ -212,6 +227,7 @@ type httpServerParameters struct {
 	trustblocAgentOpts   *trustblocAgentJSOpts
 	tls                  *tlsParameters
 	oidc                 *oidcParameters
+	keys                 *keyParameters
 }
 
 type tlsParameters struct {
@@ -225,6 +241,11 @@ type oidcParameters struct {
 	clientID     string
 	clientSecret string
 	callbackURL  string
+}
+
+type keyParameters struct {
+	sessionCookieAuthKey []byte
+	sessionCookieEncKey  []byte
 }
 
 // GetStartCmd returns the Cobra start command.
@@ -277,6 +298,11 @@ func createStartCmd(srv server) *cobra.Command {
 				return err
 			}
 
+			keys, err := getKeyParams(cmd)
+			if err != nil {
+				return err
+			}
+
 			parameters := &httpServerParameters{
 				dependencyMaxRetries: retries,
 				srv:                  srv,
@@ -286,6 +312,7 @@ func createStartCmd(srv server) *cobra.Command {
 				trustblocAgentOpts:   trustblocAgentOpts,
 				tls:                  tlsParams,
 				oidc:                 oidcParams,
+				keys:                 keys,
 			}
 
 			return startHTTPServer(parameters)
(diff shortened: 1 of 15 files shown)


cmd/http-server/startcmd/start.go
```diff
@@ -430,6 +463,56 @@ func getOIDCParams(cmd *cobra.Command) (*oidcParameters, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to configure session cooie enc key: %w", err)
+	}
+
+	return params, nil
+}
+
+func parseKey(file string) ([32]byte, error) {
+	const (
+		keyLen = 32
```
Baha-sk, line 496: I assume keyLen = 32 is because keys are used for block size 32 (AES256GCM). We should probably reference the block size of AES256GCM instead of hard coding it here.

pkg/restapi/oidc/operations.go
```diff
@@ -42,6 +46,12 @@ type Config struct {
 	Storage    *StorageConfig
 	UIEndpoint string
 	TLSConfig  *tls.Config
+	Keys       *KeyConfig
+}
+
+type KeyConfig struct {
+	Auth [32]byte
+	Enc  [32]byte
```
Baha-sk, line 54: idem for key size in type here, it should be removed

pkg/restapi/oidc/operations.go
```diff
@@ -66,9 +77,17 @@ type Operation struct {
 
 // New returns a new Operation.
 func New(config *Config) (*Operation, error) {
+	// TODO make session cookies max age configurable: https://github.com/trustbloc/edge-agent/issues/388
+	cookieStore := sessions.NewCookieStore(config.Keys.Auth[:], config.Keys.Enc[:])
```
Baha-sk, line 81: case in point about preset byte array size in the struct's fields type, the sessions api takes []byte, not [32]byte as type. This is why you had to slice Auth and Enc. You don't need to use `[:]` if the size is not preset in the type.

cmd/http-server/startcmd/start.go
```diff
@@ -227,6 +243,11 @@ type oidcParameters struct {
 	callbackURL  string
 }
 
+type keyParameters struct {
+	sessionCookieAuthKey [32]byte
```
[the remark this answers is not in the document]
llorllale: @Baha-sk `sessions.NewCookieStore()` does not fail on invalid key lengths. This means that an operator would not realize anything is wrong during startup.

Later on at runtime the service could start throwing errors like this:

```
ERROR failed to save session cookie: securecookie: error - caused by: crypto/aes: invalid key size 0
```

I think we should enforce the key lengths somehow in order to fail fast if the service is not configured properly.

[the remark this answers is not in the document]
llorllale: so I guess what you're saying is: keep the enforcement on the key length, just not encode it in the type...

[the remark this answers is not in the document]
llorllale: @Baha-sk done: removed the length from the data types but kept enforcing keys of length 32 bytes.

cmd/http-server/startcmd/start.go
llorllale: @Baha-sk where is the constant for AES256GCM block size defined?

cmd/http-server/startcmd/start.go
```diff
@@ -430,6 +463,56 @@ func getOIDCParams(cmd *cobra.Command) (*oidcParameters, error) {
...
+
+	sessionCookieEncKeyPath, err := cmdutils.GetUserSetVarFromString(cmd,
+		sessionCookieEncKeyFlagName, sessionCookieEncKeyEnvKey, false)
+	if err != nil {
+		return nil, fmt.Errorf("failed to configure session cookie enc key: %w", err)
+	}
+
+	params.sessionCookieEncKey, err = parseKey(sessionCookieEncKeyPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to configure session cooie enc key: %w", err)
```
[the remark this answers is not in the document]
llorllale: fixed

pkg/restapi/oidc/operations.go
llorllale: done

pkg/restapi/oidc/operations.go
llorllale: done

cmd/http-server/startcmd/start.go
Baha-sk: I thought I've seen AES having constant key sizes somewhere, but apparently not. So it's fine to keep it this way

Baha-sk (approved):

cmd/http-server/startcmd/start.go
```diff
@@ -169,6 +171,19 @@ const (
...
+		" Alternatively, this can be set with the following environment variable: " + sessionCookieAuthKeyEnvKey
+	sessionCookieAuthKeyEnvKey = "HTTP_SERVER_COOKIE_AUTH_KEY"
+
+	sessionCookieEncKeyFlagName  = "cookie-enc-key"
+	sessionCookieEncKeyFlagUsage = "Path to the pem-encoded 32-byte key to use to encrypt session cookies." +
+		" Alternatively, this can be set with the following environment variable: " + sessionCookieEncKeyEnvKey
+	sessionCookieEncKeyEnvKey = "HTTP_SERVER_COOKIE_ENC_KEY"
+)
+
 var logger = log.New("edge-agent/http-server")
```
troyronda, line 172: unrelated to this PR, but i think we should have a more specific binary name that makes it clear that this is the web agent / wallet server.

troyronda (approved):

cmd/http-server/startcmd/start.go
llorllale: #397

PR #389, opened 2020-10-01
fix: #381 overwriting existing user records on every login

### LINKED ISSUE #381: Create new user only if an existing user with the same OIDC `sub` does not already exist
part of #353 

PR #378 is creating a bug where a new user record is created every time new _and_ existing users login.

llorllale:
~~depends on #387~~ 

fixes #381 

### DIFF
--- a/pkg/restapi/oidc/operations.go
+++ b/pkg/restapi/oidc/operations.go
@@ -9,6 +9,7 @@ package oidc
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net/http"
 
@@ -156,9 +157,7 @@ func (o *Operation) oidcCallbackHandler(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
-	// TODO only save new user if one doesn't already exist in the store for the given `sub`:
-	//  https://github.com/trustbloc/edge-agent/issues/381
-	user := &endUser{ID: uuid.New().String()}
+	user := &endUser{}
 
 	err := user.parse(oidcToken)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/restapi/oidc/persistedData.go
```diff
@@ -61,4 +60,13 @@ func (e *persistedData) put(k string, v interface{}) error {
 	return e.s.Put(k, bits)
 }
 
+func (e *persistedData) getEndUser(sub string) (*endUser, error) {
```
aholovko, line 63: looks like that the returned `endUser` is not used, maybe we don't want to get the user but, instead, check if he exists - if this is the intention plz update the name of that function to reflect that

aholovko (approved):

pkg/restapi/oidc/persistedData.go
llorllale: The general intent is to populate the frontend with the user's name etc.

sudeshrshetty (approved):

pkg/restapi/oidc/operations.go
```diff
@@ -168,16 +167,26 @@ func (o *Operation) oidcCallbackHandler(w http.ResponseWriter, r *http.Request)
...
+		err = newPersistedData(o.store.users).put(user.Sub, user)
+		if err != nil {
+			common.WriteErrorResponsef(w, logger,
+				http.StatusInternalServerError, "failed to persist user data: %s", err.Error())
+
+			return
+		}
+	}
+
+	err = newPersistedData(o.store.tokens).put(user.Sub, &endUserTokens{
```
Baha-sk, line 188: returned `err` of put() call is not checked here, also why not use the same put() call for `err == nil` and `errors.Is(err, storage.ErrValueNotFound)` above?

llorllale: @Baha-sk 

> returned `err` of put() call is not checked here

`err` is checked right below this line.

> also why not use the same put() call for `err == nil` and `errors.Is(err, storage.ErrValueNotFound)` above?

This statement is saving `endUserTokens`. The one above is saving `endUser`.

Baha-sk: I see, it was in the unchanged code section that I haven't opened. Sounds good then.

Baha-sk (approved):

PR #436, opened 2020-10-22
feat: Uses agent-sdk

soluchok:
Uses agent-sdk.


ci/azure-pipelines.yml
```diff
@@ -50,8 +50,7 @@ jobs:
     steps:
       - template: azp-dependencies.yml
       - checkout: self
-      - script: make unit-test-wasm
-        displayName: Run unit test wasm
+      - bash: echo "no unit-test-wasm"
```
soluchok, line 53: We do not have `unit-test-wasm`. We need to remove it from `required`  then I will remove the job.

cmd/user-agent/scripts/setup_assets.sh
```diff
@@ -8,7 +8,13 @@
 rm -rf public/agent-js-worker/
 mkdir -p public/agent-js-worker/assets
 npm install
-cp -Rp node_modules/@trustbloc/agent-js-worker/dist/assets/* public/agent-js-worker/assets
+
+if [[ $(grep "@trustbloc-cicd/agent-sdk" "package.json") ]] ; then
+   cp -Rp node_modules/@trustbloc-cicd/agent-sdk/dist/assets/* public/agent-js-worker/assets
+else
+   cp -Rp node_modules/@trustbloc/agent-sdk/dist/assets/* public/agent-js-worker/assets
```
aholovko, line 15: Do we need `else` here? Under what condition it happens (don't account manual editing of `package.json`)?

soluchok: Currently, we have two different npm builds for the same repo.
@trustbloc-cicd/agent-sdk (snapshot)
@trustbloc/agent-sdk (release)

If you want to use the latest package you need `@trustbloc-cicd/agent-sdk`. 
For stable/ release you need @trustbloc/agent-sdk. 
It was done to simplify development. Otherwise, you would fix it everywhere every time when you move from `snapshot` to `release` version.

fqutishat (approved):

cmd/user-agent/scripts/setup_assets.sh
llorllale: I'm not sure how it simplifies development.

llorllale (approved):

cmd/user-agent/scripts/setup_assets.sh
soluchok: If you want to switch between versions it is enough to edit the package.json.
e.g `"@trustbloc-cicd/agent-sdk": "^0.1.5-snapshot-f8a5cb6"` or `"@trustbloc/agent-sdk": "^0.1.5"`

llorllale: Well, yes.

But updating package.json is something you'd have to do anyway.

PR #492, opened 2020-11-18
refactor: Separates UI against the backend

### LINKED ISSUE #397: `http-server` component renaming
From https://github.com/trustbloc/edge-agent/pull/387#discussion_r498565312.

We should rename the `http-server` binary (and probably the entire folder as well) that is exported to something else that captures its role and function.

Previously it was a simple static content server that served the UI's assets and offer two minimalistic endpoints to retrieve static configuration data.

#353 is now turning it into a fully-featured backend/supporting server for the frontend wallet that shares some of the workload due to security reasons:

* highly sensitive crypto material (eg. primary keys) will _never_ be hoisted unto the frontend
* likewise, access_tokens and id_tokens will never be made available to the frontend

We have also briefly mentioned that this "support" role may also be reusable for non-web wallets.

soluchok:
This PR separates UI against the backend. 
As a result, we will have two different docker images.
* `docker.pkg.github.com/trustbloc/edge-agent/user-agent-support` - backend part contains `.go` files
* `docker.pkg.github.com/trustbloc/edge-agent/user-agent` - UI part contains `.js` files 

NOTE: Since we are using the cookie when you need to test both agents you need to use user-specific domain rather than localhost. 
* `user.ui.agent.example.com`
* `second.ui.user.agent.example.com`

Can we do better?
I think so. Instead of using a cookie, we can use JWT.

Closes #397 


cmd/user-agent/src/pages/chapi/Login.vue
```diff
@@ -62,10 +62,10 @@ SPDX-License-Identifier: Apache-2.0
         },
         methods: {
             ...mapActions({loginUser: 'login', loadUser: 'loadUser', loadOIDCUser: 'loadOIDCUser'}),
-            ...mapGetters(['getCurrentUser', 'getAgentOpts']),
+            ...mapGetters(['getCurrentUser', 'getAgentOpts', 'serverURL']),
             ...mapGetters('agent', {getAgentInstance: 'getInstance'}),
             beginOIDCLogin: async function() {
-                window.location.href = "/oidc/login"
+                window.location.href = this.serverURL() + "/oidc/login"
```
llorllale, line 68: This shouldn't be necessary. The relative URL `/oidc/login` should suffice.

Hosting the frontend in a different domain from the backend will break things because the backend restricts cookies to its domain.

soluchok: @llorllale  I fixed cookies.

llorllale: @soluchok they didn't need fixing.

test/bdd/fixtures/agent-wasm/docker-compose.yml
```diff
@@ -125,18 +146,29 @@ services:
...
+      - AGENT_DEFAULT_LABEL=user-agent
+      - BLOC_DOMAIN=testnet.trustbloc.local
+      - EDGE_AGENT_SERVER=https://localhost:8077
+      - WALLET_MEDIATOR_URL=${WALLET_ROUTER_URL}
+
   bddtest.user.agent.example.com:
     container_name: bddtest.user.agent.example.com
     image: ${USER_WASM_IMAGE}:latest
     environment:
+      - AGENT_UI_URL=https://localhost:8078
```
rolsonquadras, line 167: @soluchok Looks like we have cyclic dependency - agent server calling web-wallet-ui and [vice versa](https://github.com/trustbloc/edge-agent/pull/492/files#diff-0325567f17eab8064d99ba9a4ac40c08a40d423010c14d597c0ab8eae6e62ce1R160).

cmd/user-agent/src/pages/chapi/Login.vue
soluchok: It is necessary in case if we have the same domain but different ports.

llorllale: https://tools.ietf.org/html/rfc6265#section-8.5

Good finding @soluchok

docs/test/test.md
```diff
@@ -60,8 +67,8 @@ make user-agent-start
 
 ## Agents
 
-- To access user agent wasm open [user home page](https://localhost:8091/wallet/dashboard).
-- To access second user agent wasm open [user home page](https://localhost:8071/wallet/dashboard).
+- To access user agent wasm open [user home page](https://user.ui.agent.example.com:8091/dashboard).
+- To access second user agent wasm open [user home page](https://second.ui.user.agent.example.com:8071/dashboard).
```
llorllale, line 71: We should add a note here saying the user needs to open the second agent in a different browser (not just tab) so the login will work as expected after #497 is fixed.

If their browser supports containers, they can use the same browser but open the login page in different containers.

llorllale (approved):

ci/version_var.sh
```diff
@@ -11,7 +11,9 @@ BASE_VERSION=0.1.5
 IS_RELEASE=false
 
 SOURCE_REPO=edge-agent
-BASE_USER_WASM_PKG_NAME=user-agent-wasm
+BASE_USER_AGENT_PKG_NAME=user-agent
+BASE_USER_AGENT_SUPPORT_PKG_NAME=user-agent-support
```
rolsonquadras, line 15: IMO web-wallet/agent and web-wallet/agent-support would be more appropriate names for the images, as these are just supporting agent features on browser/web.

@troyronda @fqutishat @llorllale @soluchok thoughts ?

docs/test/test.md
soluchok: I will add it.

rolsonquadras (approved):

ci/version_var.sh
troyronda: Other wallet types also need recovery via login. It's not clear that _all_ the features are limited to a Web Wallet.

troyronda: I do think user-agent is a confusing name for a backend service.

PR #501, opened 2020-11-20
refactor: Rename SDS to EDV and remove legacy EDV storage code

### LINKED ISSUE #328: Deal with SDS sync failures better
Right now whenever a DID, credential, or presentation is saved, these two things happen in this order:

1. Save to local storage
2. Save to SDS

What if saving to SDS fails? The document will still be saved locally, but it won't be in the SDS... and never will. How should we deal with this? Some sort of background service that ensures the documents are synced up?

DRK3:
- Renamed SDS to EDV to align with the current naming scheme used in aries-framework-go and agent-sdk.
- Removed the old EDV storage code which isn't functional anymore since it was removed from agent-sdk already. In a future commit, EDV storage will be handled by agent-sdk.
- Added missing declaration for the OPS_EDV_PORT environment variable in the .env file.

closes #328 since it's no longer applicable with the changes in this commit.


Baha-sk (approved):

aholovko (approved):

test/bdd/fixtures/user-agent/.env
```diff
@@ -73,3 +73,4 @@ HUB_KMS_IMAGE_TAG=0.1.5-snapshot-981f6da
 HUB_KMS_HOST=0.0.0.0
 AUTHZ_HUB_KMS_PORT=8076
 OPS_HUB_KMS_PORT=8075
+OPS_EDV_PORT=8072
```
rolsonquadras, line 76: This is not required. We have already declared variable for `8072` on line 59.

Issue #506, opened 2020-11-20
Use agent-sdk's EDV storage option


DRK3:

DRK3:
Not sure which PR closed this issue, but this was completed a while back

PR #508, opened 2020-11-20
feat: EDV storage

DRK3:

### DIFF
--- a/cmd/user-agent/src/store/modules/options.js
+++ b/cmd/user-agent/src/store/modules/options.js
@@ -25,8 +25,9 @@ let defaultAgentStartupOpts = {
     walletMediatorURL: 'https://localhost:10063',
     blindedRouting: false,
     credentialMediatorURL: '',
-    storageType: `indexedDB`, 	// TODO (#475): Allow the storage type to be configurable.
-    edvServerURL: ''
+    storageType: `edv`,
+    edvServerURL: '',
+    edvVaultID: ''
 }
 
 export default {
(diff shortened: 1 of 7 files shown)


images/user-agent/Dockerfile
```diff
@@ -1,3 +1,7 @@
+# Copyright SecureKey Technologies Inc. All Rights Reserved.
```
DRK3, line 1: For some reason locally the license text check fails for me on this file now. The CI doesn't seem to have a problem, though.

llorllale:
@DRK3 does this close #506?

cmd/user-agent/src/store/modules/options.js
```diff
@@ -50,6 +51,31 @@ export default {
...
+
+                console.log("User info URL is: " + userInfoURL)
+
+                const client = axios.create({
+                    withCredentials: true
+                })
+
+                await client.get(userInfoURL)
+                    .then(resp => {
+                        const edvVaultURL = resp.data.bootstrap.edvVaultURL
```
llorllale, line 65: We're going to need alignment at some point w.r.t. the bootstrap "URLs" (values the agent cannot change at hub-auth) vs. the vault and keystore IDs.

I keep seeing usage of what are supposed to be just "vault/keystore IDs" being used to store the full URL.

cc @Baha-sk

llorllale (approved):

llorllale:
@DRK3 can you confirm the [tests](https://github.com/trustbloc/edge-agent/blob/master/docs/test/test.md) work with this change?

fqutishat (approved):

cmd/user-agent/src/store/modules/options.js
Baha-sk: vaultID, keystoreID are supposed to be the full URL to the vault/keystore @llorllale 

The user should not be able to modify this value after onboarding. Only admins can update these values for migration purposes.

cc @DRK3 @aholovko @rolsonquadras

cmd/user-agent/src/store/modules/options.js
```diff
@@ -50,6 +51,31 @@ export default {
...
+
+                        console.log("User EDV Vault URL is: " + edvVaultURL)
+
+                        const edvVaultID = edvVaultURL.substring(edvVaultURL.lastIndexOf('/')+1)
+
+                        console.log("User EDV Vault ID is: " + edvVaultID)
```
Baha-sk, line 71: @DRK3 let's remove vaultID.. only vaultURL is needed in hub-auth.

hub-auth should have only 1 value for edv vault (and only 1 value for the keystore). 

cc @llorllale , @aholovko @rolsonquadras @fqutishat

DRK3: This `substring` code just extracts the vault ID from the vault URL. The vault ID is used by the Aries EDV REST provider. We could potentially make hub-auth just store the vault ID instead of the full URL. I don't think it makes much difference either way, though.

PR #511, opened 2020-11-22
fix: Updated EDV server docker image

DRK3:
- The updated EDV server fixes an issue that would cause storage failures once the number of stored documents reached a certain threshold.
- Made EDV storage the default storage type.


DRK3:
I tested all flows in https://github.com/trustbloc/edge-agent/blob/master/docs/test/test.md tested and confirmed that they work.

fqutishat (approved):

PR #519, opened 2020-11-24
feat: EDV vault ID storage workaround

DRK3:
A temporary workaround to allow the user's EDV vault ID to be accessed from a CHAPI window.

### DIFF
--- a/cmd/wallet-web/src/store/modules/options.js
+++ b/cmd/wallet-web/src/store/modules/options.js
@@ -4,6 +4,8 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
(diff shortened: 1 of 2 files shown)


fqutishat (changes requested):
before we merge let's discuss how long will take to fix the original issue

fqutishat:
this workaround will not work for edv zcap

llorllale (approved):

troyronda (commented):
We need to be able to switch browsers to complete this milestone. So we need to get vault IDs from the auth service. I’m not entirely sure there is a point in this workaround as it need to be fixed ASAP?

fqutishat (approved):

DRK3:
@troyronda This is only a temporary workaround. Note that hub-auth still stores the vault ID for recovery purposes - this change by itself doesn't prevent switching browsers.

PR #522, opened 2020-11-24
fix: Fixes cookie usage in iframe

soluchok:
This PR fixes:
* cookie usage in iframe
* cors subdomain issue


fqutishat (approved):

talwinder50 (approved):

llorllale (approved):
Should work for now since cookies are now shared on the same subdomain. I think the mechanism should be generalized later on to allow SSO with disparate domains.

Issue #488, opened 2020-11-17
Update keystore request to pass the half secret to hub-kms


rolsonquadras:
Related to https://github.com/trustbloc/edge-agent/issues/485

PR #539, opened 2020-11-27
feat: Pass half secret share to hub-kms

rolsonquadras:
part #488 

Signed-off-by: Rolson Quadras <rolson.quadras@securekey.com>

### DIFF
--- a/pkg/restapi/oidc/operations.go
+++ b/pkg/restapi/oidc/operations.go
@@ -247,13 +247,15 @@ func (o *Operation) oidcCallbackHandler(w http.ResponseWriter, r *http.Request) 
 	}
 
 	if errors.Is(err, storage.ErrValueNotFound) {
-		err = o.onboardUser(usr.Sub, oauthToken.AccessToken)
-		if err != nil {
+		walletSecretShare, onboardErr := o.onboardUser(usr.Sub, oauthToken.AccessToken)
+		if onboardErr != nil {
 			common.WriteErrorResponsef(w, logger,
-				http.StatusInternalServerError, "failed to onboard the user: %s", err.Error())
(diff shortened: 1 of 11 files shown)


fqutishat (approved):

pkg/restapi/oidc/operations.go
```diff
@@ -507,64 +521,70 @@ func (o *Operation) userLogoutHandler(w http.ResponseWriter, r *http.Request) {
...
 	secrets, err := o.secretSplitter.Split(b, 2, 2)
 	if err != nil {
-		return fmt.Errorf("split user secret key : %w", err)
+		return "", fmt.Errorf("split user secret key : %w", err)
 	}
 
-	// TODO https://github.com/trustbloc/edge-agent/issues/488 send half secret key to hub-kms and remove logger
-	logger.Infof(string(secrets[1]))
+	walletSecretShare := base64.StdEncoding.EncodeToString(secrets[0])
+	hubAuthSecretShare := secrets[1]
```
aholovko, line 538: @Baha does it matter in what order we are combining secrets? Currently Key Server expects to get secret A in the header and fetch secret B from hub-auth. Here is the opposite situation. We probably should agree on order.

aholovko (changes requested):

talwinder50 (approved):

pkg/restapi/oidc/operations.go
rolsonquadras: @aholovko As far as I understand, the order doesn't matter and I was able to create keys locally (for now I have disabled the hub-auth integration as I am facing other issues).

aholovko (approved):

PR #597, opened 2020-12-07
fix: Login UX change and alert fixes.

talwinder50:
1. Redesign Login Page
2. Show Login/Register button for non-registered user. 
3. Pop up Login with Touch ID option for the registered user. 
4. Show the register Device button only for the non-registered device. 
5. Image resolution is reduced 

Screen Shot 2020-12-07 at 3 49 31 PM

### DIFF
--- a/cmd/wallet-web/src/pages/chapi/Login.vue
+++ b/cmd/wallet-web/src/pages/chapi/Login.vue
@@ -5,12 +5,15 @@ SPDX-License-Identifier: Apache-2.0
 */
 
 <template>
-    <div>
-        <md-card style="width: 20%; margin: 15% 0% 0% 40%" md-with-hover>
+    <div class="viewport">
+        <md-card class="md-layout-item md-size-50 md-small-size-100">
             <md-card-header data-background-color="green">
-                <div class="md-title">Login to your wallet</div>
+                <div class="md-title text-center">Web Wallet Login</div>
             </md-card-header>
-
+            <md-card-content>
+                <img src="@/assets/img/wallet.jpg">
+            </md-card-content>
+            <div class="md-body-1 center-header">Log in or register with the same sign-in information you use for other online services (for example, online banking).</div>
             <md-card-content v-if="loading" style="margin: 10% 22% 10% 30%">
                 <md-progress-spinner :md-diameter="100" class="md-accent" :md-stroke="10"
                                      md-mode="indeterminate"></md-progress-spinner>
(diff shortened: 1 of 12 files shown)


rolsonquadras (approved):

cmd/wallet-web/src/pages/chapi/Login.vue
```diff
@@ -36,13 +39,16 @@ SPDX-License-Identifier: Apache-2.0
 <script>
     import {DeviceLogin, RegisterWallet} from "./wallet"
     import {mapActions, mapGetters} from 'vuex'
+    import Vue from 'vue';
 
     export default {
         created: async function () {
+           this.registerSuccess = Vue.$cookies.get("device");
```
llorllale, line 46: Cookies should not be set/read by the frontend :(.

cmd/wallet-web/src/pages/chapi/Login.vue
```diff
@@ -111,8 +121,36 @@ SPDX-License-Identifier: Apache-2.0
             loginDevice: async function () {
                 await this.deviceLogin.login();
             },
+            getCookie (name) {  /// Retrieve cookie function
+               var match = document.cookie.match(new RegExp(name + '=([^;]+)'));
```
llorllale, line 125: Why is the regexp needed?

cmd/wallet-web/src/pages/chapi/Login.vue
talwinder50: This cookie dont have sensitive data it is just to enable / disable  login with touch id button in the front end. Now reading this value only enabling Sign in with touch id option if the device has been registered. I can add todo to re-visit, but in my opinion it should be fine.

llorllale: @talwinder50 if the user wipes their browser's storage (re-install, etc) then the "Login With Touch ID" button would not be enabled, correct?

This would be better implemented with information from the backend.

I guess that's a TODO.

cmd/wallet-web/src/pages/chapi/Login.vue
talwinder50: this is no longer required, missed in the cleanup. removed now.

cmd/wallet-web/src/pages/chapi/Login.vue
talwinder50: add #601 as todo

llorllale (approved):

PR #626, opened 2021-01-04
feat: Create card like view for stored credentials

talwinder50:

### DIFF
--- a/cmd/wallet-web/src/components/CredentialCards/GeneralCard.vue
+++ b/cmd/wallet-web/src/components/CredentialCards/GeneralCard.vue
@@ -1,0 +1,29 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+<template>
+    <div class="cardBody">
+        <div class="cardDetailsL">
+            <md-icon>perm_identity</md-icon>
+        </div>
+        <div class="cardDetailsR">
+            <p> {{ item.description || 'Verifiable Credential'}}</p>
+        </div>
+    </div>
+</template>
+
+<script>
+    export default {
+        name: "general-card",
+        props: {
+            item: Object,
+        },
+    };
+</script>
+
+<style lang="scss">
+    @import "@/assets/scss/card-content";
+</style>
--- a/cmd/wallet-web/src/pages/Dashboard.vue
+++ b/cmd/wallet-web/src/pages/Dashboard.vue
@@ -30,13 +30,57 @@ SPDX-License-Identifier: Apache-2.0
                             </h4>
                         </md-card-header>
                         <md-card-content>
-                            <simple-table v-for="vc in verifiableCredentials"
-                                          :key=vc.id
-                                          :name=credDisplayName(vc)
-                                          :description=vc.description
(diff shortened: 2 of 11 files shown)


fqutishat (approved):

cmd/wallet-web/src/components/CredentialCards/GeneralCard.vue
```diff
@@ -0,0 +1,29 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+<template>
```
sudeshrshetty, line 7: is this for credentials which doesn't fall under category?

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -127,6 +194,74 @@ SPDX-License-Identifier: Apache-2.0
         font-size: 6px;
         line-height: 16px;
     }
+    ul.credential-list {
+        padding-left: 0;
+        display: flex;
+        flex-flow: row wrap;
+    }
+
+    li {
```
soluchok, line 203: you are changing `li` which is global, but the `<style>` is not `scoped`.

sudeshrshetty (commented):
follow up: 
Can we have just one generic vue card for all the credentials?
we can use vue props to pass icons & credential subject properties.

So that, it can be dynamically updated without code change to add support more VC types.

talwinder50:
Okay i will look into that approach .. do you want me to create todo or do in this pr only ?

talwinder50:
> follow up: 
> 
> Can we have just one generic vue card for all the credentials?
> 
> we can use vue props to pass icons & credential subject properties.
> 
> 
> 
> So that, it can be dynamically updated without code change to add support more VC types.  

Okay i will look into that approach .. do you want me to create todo or do in this pr only

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: Will make style scoped

sudeshrshetty:
> > follow up:
> > Can we have just one generic vue card for all the credentials?
> > we can use vue props to pass icons & credential subject properties.
> > So that, it can be dynamically updated without code change to add support more VC types.
> 
> Okay i will look into that approach .. do you want me to create todo or do in this pr only

new PR should be fine.

cmd/wallet-web/src/components/CredentialCards/GeneralCard.vue
talwinder50: yes

talwinder50:
> > > follow up:
> > > Can we have just one generic vue card for all the credentials?
> > > we can use vue props to pass icons & credential subject properties.
> > > So that, it can be dynamically updated without code change to add support more VC types.
> > 
> > 
> > Okay i will look into that approach .. do you want me to create todo or do in this pr only
> 
> new PR should be fine.

created todo #627

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: done

soluchok (approved):

PR #720, opened 2021-05-11
feat: Introduce tailwind css and new designed landing page

talwinder50:
Screen Shot 2021-05-11 at 5 32 40 PM

Screen Shot 2021-05-11 at 5 27 20 PM


rolsonquadras:
@talwinder50 What does login and sign-up button do ? Currently backend doesn't support different options. If user is not there then it signs-up by default.

cmd/wallet-web/src/main.js
```diff
@@ -11,9 +11,10 @@ import VueRouter from "vue-router";
 import routes from "./router/index";
 import * as polyfill from "credential-handler-polyfill";
 import * as webCredentialHandler from "web-credential-handler";
-import MaterialDashboard from "./material-dashboard";
+//import MaterialDashboard from "./material-dashboard";
```
sudeshrshetty, line 14: why comment? why can't remove it?

talwinder50: it was a typo, missed to roll back. Done now.

fqutishat (approved):

PR #904, opened 2021-08-06
feat: added eslint and prettier checks to CI

### LINKED ISSUE #900: Add prettier and eslint checks to CI

birtony:
Fixes #900

Signed-off-by: Anton Biriukov <anton.biriukov@securekey.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,18 +4,50 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-name: "wallet ci"
+name: "Wallet CI"
 
 env:
   GO_VERSION: 1.16
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
 
 jobs:
+  Prettier:
+    name: Prettier Check
+    timeout-minutes: 5
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run Wallet Web Prettier Check
+        run: |
+          echo '//npm.pkg.github.com/:_authToken=${GITHUB_TOKEN}' > ~/.npmrc
+          echo '@hyperledger:registry=https://npm.pkg.github.com' >> ~/.npmrc
+          echo '@trustbloc-cicd:registry=https://npm.pkg.github.com' >> ~/.npmrc
+          echo '@trustbloc:registry=https://npm.pkg.github.com' >> ~/.npmrc
+          make wallet-web-prettier-check
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  ESLint:
+    name: ESLint Check
+    timeout-minutes: 5
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -4,18 +4,62 @@
...
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Run Wallet Web Prettier Check
+        run: |
+          echo '127.0.0.1 testnet.orb.local' | sudo tee -a /etc/hosts
```
sudeshrshetty, line 31: Are you sure we need these hostfile entries?

.github/workflows/build.yml
```diff
@@ -4,18 +4,62 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  ESLint:
+    name: ESLint Check
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Go
```
sudeshrshetty, line 47: Setup Go?

.github/workflows/build.yml
birtony: Not sure, should I try without them?

rolsonquadras: We don;t need the host file - its only needed when we start docker and run

sudeshrshetty: yes, those are used for bddtests to avoid connection errors between containers

.github/workflows/build.yml
```diff
@@ -145,7 +189,7 @@ jobs:
         with:
           current-version: "${{ env.CURRENT_SEMVER }}"
           # TODO: needs to be configurable https://github.com/christian-draeger/increment-semantic-version#input--output-examples
-          version-fragment: 'bug'
+          version-fragment: "bug"
```
rolsonquadras, line 192: Do we need this change ?

birtony: Is there any specific reason why these quotes are single while the rest of them are double?

.github/workflows/build.yml
birtony: Ok, thanks, removed

.github/workflows/build.yml
birtony: Can you elaborate, please? @sudeshrshetty

.github/workflows/build.yml
rolsonquadras: https://github.com/christian-draeger/increment-semantic-version#example-usage

rolsonquadras: This is fine - I was just curious.

.github/workflows/build.yml
rolsonquadras: I believe, @sudeshrshetty meant do we need this step ?

rolsonquadras (approved):

.github/workflows/build.yml
birtony: I see, we probably don't - removed

sudeshrshetty (approved):

Issue #914, opened 2021-08-09
After loading third party login , loader dont get removed. 


talwinder50:
the loader if user closes the window while its redirecting, but when its redirected, the third party then loader dont get removed

PR #915, opened 2021-08-09
feat: Integrate toast notification for system failures and remove loader on closed popup

### LINKED ISSUE #898: Integrate toast component at the server failure in signup /sign in flow

### LINKED ISSUE #899: Display list of provider if pop up login is closed without logging in

talwinder50:
The toast notification is tested with adding wrong param in option.js file to simulate system failure. 

One existing issue. Pop up it self removes the loader if user closes the window while its redirecting, but when its redirected, the third party then loader dont get removed. I havent found a way where window.listener can unload from window.location.href.  Issue #914 is created for that

https://user-images.githubusercontent.com/7862595/128655697-8331f83c-6dab-4cf3-ac6c-59d2c026ff80.mov

closes #899 
closes #898

Signed-off-by: talwinder kaur <talwinder.kaur@securekey.com>

### DIFF
--- a/cmd/wallet-web/src/pages/Signin.vue
+++ b/cmd/wallet-web/src/pages/Signin.vue
@@ -18,6 +18,12 @@
   >
     <div class="flex flex-col flex-grow justify-center items-center">
       <!--Trustbloc Sign-up provider div -->
+      <toast-notification
+        v-if="systemError"
+        :title="i18n.errorToast.title"
+        :description="i18n.errorToast.description"
+        type="error"
+      ></toast-notification>
       <div
         class="
           overflow-hidden
@@ -106,6 +112,7 @@ export default {
       providers: [],
       statusMsg: '',
       loading: true,
+      systemError: false,
     };
   },
   computed: {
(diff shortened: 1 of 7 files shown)


cmd/wallet-web/src/pages/Signin.vue
```diff
@@ -18,6 +18,12 @@
   >
     <div class="flex flex-col flex-grow justify-center items-center">
       <!--Trustbloc Sign-up provider div -->
+      <toast-notification
+        v-if="systemError"
+        :title="i18n.errorToast.title"
+        :description="i18n.errorToast.description"
```
birtony, line 24: Translations won't get loaded. There is no `errorToast` field in the `src/translations/en|fr.json` for the `Signin` screen.

cmd/wallet-web/src/pages/Signin.vue
```diff
@@ -18,6 +18,12 @@
...
+        type="error"
+      ></toast-notification>
```
birtony, line 26: What do you think about using camel case notation for user-defined element? For instance, `<ToastNotification />` instead of what we have right now. To me, it makes to easier to understand that this component is custom.

cmd/wallet-web/src/pages/Signup.vue
```diff
@@ -148,20 +155,30 @@ export default {
     isLoggedIn() {
       return this.isUserLoggedIn();
     },
+    isSuspended() {
+      return this.isLogInSuspended();
```
birtony, line 159: `isLoginSuspended` would be more precise, as `login` is a noun here

birtony (changes requested):

cmd/wallet-web/src/pages/Signin.vue
talwinder50: translation files didnt get pushed earlier. Thanks updated

cmd/wallet-web/src/pages/Signup.vue
talwinder50: updated

cmd/wallet-web/src/pages/Signin.vue
talwinder50: https://v3.vuejs.org/guide/component-registration.html#component-names

birtony (approved):

rolsonquadras (approved):

Issue #979, opened 2021-08-19
Can't see credential in the list view


rolsonquadras:
Saved Flight booking reference credential from demo-issuer, but its not listed in list-view.

Issue #974, opened 2021-08-19
Lazy-load credential icons 


talwinder50:
We should lazy-load images preferably, instead of bulk-loading them

_Originally posted by @birtony in https://github.com/trustbloc/wallet/pull/973#discussion_r691707136_

PR #1005, opened 2021-08-27
feat: Refactor credential preview json

### LINKED ISSUE #994: Refactor credential detail json file and and handle generic credential

talwinder50:
closes #994  #979  #974

I have tried deleting and splicing the credential from the original array but it behaves weirdly so i opted for the following solution:

1. We look for the key type in the credentials saved and check if the credential type exists in the manifest json config file, if yes that will be treated as manifested credential , otherwise it will be treated as generic credential, and object type = `generic `or `manifest `will be added to the `processedcredential ` array and we can render based on the type.

Following videos show , simple example, if we have the credential manifest available or not based on that preview gets changes, notice the icon change. 

https://user-images.githubusercontent.com/7862595/131265984-b9482006-c860-40f0-8a5c-ea15563ba248.mov

In this scenario PR Card and Vaccination credentials types are present in the config file and Booking ref is not still we see it shows in the preview. 

https://user-images.githubusercontent.com/7862595/131266234-10544eb1-738a-4766-8d3a-b433a860d6df.mov


birtony:
After the discussion with @talwinder50, we have come up with the following logic on how to process and display credentials:

1. We have two objects - `vc` (the one we get from whatever user saves to wallet vie web-wallet page) and `manifest` (`credentialDisplayData.json` - the types of objects that we want to display in special styles)
2. When the credentials view page is rendered, we need to decide which credentials to render and what styles to apply (general vs supported)
3. We will push all manifested credentials in a separate array, called `manifestedCredentials`, while general credentials will be pushed to `generalCredentials` array.
4. In order to determine whether a credential is manifested, we need to iterate over the `vc` object and check the credential type by calling the `getCredentialType(vc.type)` helper function from `helper.js` and verify whether such type exists in `credentialDisplayData.json`.
5. If the type exists, then such credential will be treated as manifested credential. We need to iterate through it and call `jp.query` function to replace all JSON paths (i.e. queries in any `path` property within credential object). Afterwards, we push the credential object to `manifestedCredentials` array. If the type does not exists, we push the credential object to `generalCredentials`.
6. When rendering credential previews, we need to check `manifestedCredentials` and `generalCredentials` arrays via `v-if` clause. If the length is 0, then no credential previews are displayed for such credential type. If both of the arrays have 0 length, then the no credential container is rendered.

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,19 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+  let updatedManifestCredential;
+  Object.keys(manifestCredential).forEach(function (key) {
+    updatedManifestCredential = JSON.stringify(
+      manifestCredential,
+      (k, v) => (k == 'path' ? jp.query(vc, v.toString()) : v),
+      1
+    );
+  });
+  return JSON.parse(updatedManifestCredential);
+}
```
birtony, line 103: Can we do something like this?
```js
export function updateCredentialJsonPath(vc, manifestCredential) {
  const updatedManifestCredential = Object.keys(manifestCredential).forEach((key) =>
    JSON.stringify(
      manifestCredential,
      (k, v) => (k == 'path' ? jp.query(vc, v.toString()) : v),
      1
    );
  });
  return JSON.parse(updatedManifestCredential);
}
```

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -139,42 +165,46 @@ export default {
...
-          for (let credentialContent in flattened) {
-            if (
-              obj.credentialSubject.hasOwnProperty(credentialContent) &&
-              obj.schema === this.credDisplayName(this.cards[credentialKey].content)
-            ) {
-              obj.credentialSubject[credentialContent] = flattened[credentialContent];
-            }
+      Object.entries(this.credentials).forEach(([key]) => {
+        let manifestCredential;
+        let cType = this.filterType(this.credentials[key].content);
```
birtony, line 174: If we don't plan to modify this variable, I would suggest making it a `const` - see https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75 for reference, for instance

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -104,8 +128,10 @@ export default {
   },
   data() {
     return {
-      cards: [],
-      cjson: credentialDisplayData,
+      credentials: [],
+      genericCredentials: [],
+      manifestedCredentials: [],
```
birtony, line 133: I am just thinking if we need two separate arrays after all?

cmd/wallet-web/src/config/credentialDisplayData.json
```diff
@@ -1,105 +1,268 @@
...
     },
-    "icon": "credential--generic-icon.svg"
+    "icon": "credential--generic-icon.svg",
+    "title": {
+      "path": ["$.name"],
+      "fallback": "Student Card"
+    }
   }
-]
+}
```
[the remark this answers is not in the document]
talwinder50: Prettier complaints and fails so i have ran prettier

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: Yes for now see the description i have tried and already spent so much time to figure out to delete or splice the original array this is the simpler solution for now.

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -139,42 +165,46 @@ export default {
         return { content: contents[id], flipped: false };
       };
 
-      this.cards = Object.keys(contents).filter(_filter).map(_createCard);
+      this.credentials = Object.keys(contents).filter(_filter).map(_createCard);
 
-      console.log(`showing ${this.cards.length} credentials`);
+      console.log(`showing ${this.credentials.length} credentials`);
```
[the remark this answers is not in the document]
talwinder50: We need this for debugging purposes. Till we have logger setup. @sudeshrshetty what you think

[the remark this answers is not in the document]
birtony: At least use `console.debug`

cmd/wallet-web/src/pages/Dashboard.vue
birtony: No, I'm not saying we need to modify the original array. I'm thinking if we should just push both general and manifested credentials to the same array instead splitting them up in two?

talwinder50: I dont see how to to do that we need to render both general and manifested credential at the same time and both have different thing to render on ui

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: Unfortunately we have plan to modify this variable and using const dont work , i needed to use let.

cmd/wallet-web/src/pages/mixins/common/helper.js
talwinder50: const leads to the javascript error for this scenario.

cmd/wallet-web/src/pages/Dashboard.vue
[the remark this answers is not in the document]
talwinder50: done

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: we need to render different ui container for the two scenario manifest and general, pushing them to the same array will not allow us to do that

cmd/wallet-web/src/pages/mixins/common/helper.js
birtony: What error?

cmd/wallet-web/src/pages/Dashboard.vue
birtony: But you don't modify it anywhere

cmd/wallet-web/src/pages/Dashboard.vue
birtony: In the template, we could check the type of the credential from the array and render corresponding UI

cmd/wallet-web/src/config/credentialDisplayData.json
```diff
@@ -1,105 +1,268 @@
-[
-  {
-    "schema": "Bachelor Degree",
-    "credentialSubject": {
-      "description": ""
+{
+  "PermanentResidentCard": {
+    "properties": {
+      "birthCountry": {
+        "path": "$.credentialSubject.birthCountry",
```
sudeshrshetty, line 5: path has to be array as per standards according to various specs

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -43,13 +43,36 @@
...
         <span class="font-bold font-xl text-neutals-dark">{{ i18n.defaultvault }}</span>
       </div>
       <ul class="grid grid-cols-1 md:grid-cols-2 gap-4 md:gap-8 mx-6 md:mx-0">
-        <li v-for="(card, index) in cards" :key="index">
-          <!--Temporary solution to handle preview for the generic credential, this will be refactored issue-981-->
+        <li v-for="(manifestCredential, index) in manifestedCredentials" :key="index">
+          <div class="inline-flex items-center w-full credentialCard">
+            <div class="flex-none w-12 h-12 border-opacity-10">
+              <!-- Todo lazy load icons #974 -->
+              <img :src="loadImg(manifestCredential.icon)" />
```
sudeshrshetty: why hardcoding manifest credential?

sudeshrshetty (changes requested):

cmd/wallet-web/src/config/credentialDisplayData.json
```diff
@@ -1,105 +1,268 @@
...
+        "path": "$.credentialSubject.issue_date",
+        "label": "Date of Issue",
+        "type": ""
+      }
     },
-    "icon": "credential--generic-icon.svg"
+    "icon": "credential--generic-icon.svg",
+    "title": {
+      "path": ["$.name"],
+      "fallback": "Student Card"
```
rolsonquadras, line 265: This doesn't look right for generic case - falling back to 'Student Card'

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,52 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+// function to replace json path from the json objects
+function findByPath(vc, ...paths) {
+  let pathValue;
+  paths.forEach((path) => {
+    path.forEach((nestedPath) => {
+      pathValue = jp.query(vc, nestedPath);
+    });
+  });
+  // returns the first value which is path[0].
+  return pathValue.shift();
```
birtony, line 135: Why do you need to iterate two times and call a jp.query function on each iteration? This is very expensive and does not seem right

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,52 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+  const titleFound = findByPath(vc, manifest.title.path);
+  const title = titleFound ? titleFound : manifest.title.fallback;
+
+  // icon
+  const icon = manifest.icon;
+
+  // find properties
+  const _readProperty = (property) => {
+    let pKey = property[0];
+    let vals = findByPath(vc, property[1].path);
```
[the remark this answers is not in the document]
talwinder50: It doesn't work with const.. i have already tried n spent much time.

cmd/wallet-web/src/pages/mixins/common/helper.js
talwinder50: updated to single for loop, thanks

cmd/wallet-web/src/pages/mixins/common/helper.js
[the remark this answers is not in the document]
talwinder50: done

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,50 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+  const properties = Object.entries(manifest.properties).map(_readProperty);
+  return {
+    title,
+    icon,
+    properties,
+  };
+}
+// function to replace json path from the json objects
+function findByPath(vc, ...paths) {
+  let pathValue;
```
birtony, line 128: Make this `const`

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,50 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+  for (const key in paths) {
+    pathValue = jp.query(vc, paths[key][0]);
+  }
+  // returns the first value which is path[0].
+  return pathValue.shift();
```
birtony, line 133: Just return `pathValue[0]` - you don't need to modify `pathValue` array

birtony: I'm still not clear about this function. We should only call `jp.query` on `path` properties, not on every path. Besides, you only pass single string to `paths` on lines https://github.com/trustbloc/wallet/pull/1005/files#diff-901fbc10d64331f455b4ac2aebc1f882cb8a5fded27e30dadc624a917e75ec1cR104 and https://github.com/trustbloc/wallet/pull/1005/files#diff-901fbc10d64331f455b4ac2aebc1f882cb8a5fded27e30dadc624a917e75ec1cR95 - so what's the type of `paths`? Should we rename it to `path` if it's just a single string?

cmd/wallet-web/src/pages/mixins/common/helper.js
talwinder50: val is read only error comes, i already spent time and it didnt work for me

sudeshrshetty: you have to return on very first match, no need define a variable outside loop

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,50 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
 export const getCredentialType = (types) =>
   types.filter((type) => type != 'VerifiableCredential')[0];
 
+// function to replace json path from the json objects
```
sudeshrshetty, line 92: not a function to replace

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,50 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+  const title = titleFound ? titleFound : manifest.title.fallback;
+
+  // icon
+  const icon = manifest.icon;
+
+  // find properties
+  const _readProperty = (property) => {
+    const pKey = property[0];
+    const vals = findByPath(vc, property[1].path);
+    const value = vals ? vals : '';
```
birtony, line 105:
```suggestion
    const value = findByPath(vc, property[1].path) || '';
```

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,50 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+      [pKey]: {
+        label,
+        value,
+        type,
+        format,
+      },
+    };
+  };
+
+  const properties = Object.entries(manifest.properties).map(_readProperty);
```
sudeshrshetty, line 117: Need a default logic for credential type not found in manifest, you need a default manifest entry. and in those cases you have to display all properties.

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,50 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+  // find properties
+  const _readProperty = (property) => {
+    const pKey = property[0];
+    const vals = findByPath(vc, property[1].path);
+    const value = vals ? vals : '';
+    const label = property.label;
+    const type = property.type;
+    const format = property.format;
+    return {
+      [pKey]: {
```
sudeshrshetty, line 108: why you need `[pKey]` here??

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -139,42 +161,55 @@ export default {
         return { content: contents[id], flipped: false };
       };
 
-      this.cards = Object.keys(contents).filter(_filter).map(_createCard);
+      this.credentials = Object.keys(contents).filter(_filter).map(_createCard);
```
sudeshrshetty, line 164: Why you need `_createCard` still?

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -139,42 +161,55 @@ export default {
...
-          this.flatten(this.cards[credentialKey].content, flattened);
-          for (let credentialContent in flattened) {
-            if (
-              obj.credentialSubject.hasOwnProperty(credentialContent) &&
-              obj.schema === this.credDisplayName(this.cards[credentialKey].content)
-            ) {
-              obj.credentialSubject[credentialContent] = flattened[credentialContent];
-            }
+      Object.entries(this.credentials).forEach(([key]) => {
+        const cType = this.filterType(this.credentials[key].content);
```
sudeshrshetty, line 169: You looped on object entries and extract key inside loop ? `[key]`?
credential should be function arg here.

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -139,42 +161,55 @@ export default {
...
+        Object.entries(this.manifestCredJson).forEach(([credType]) => {
```
sudeshrshetty, line 170: why you need one more loop here??
it should this.manifestCredJson[credType]

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -139,42 +161,55 @@ export default {
...
+            Object.defineProperties(processedCredential, {
+              type: {
+                value: 'manifested',
+              },
+            });
+            this.processedCredentials.push(processedCredential);
           }
+        });
+        if (!this.manifestCredJson.hasOwnProperty(cType)) {
+          const genericCredential = this.credentials[key];
```
sudeshrshetty, line 185: this whole logic is being over complicated, it can be way simpler

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -139,42 +161,55 @@ export default {
...
-          }
-        }
+    replacer: function (key, value) {
+      if (typeof value === 'function') {
+        return value.toString();
       }
+      return value;
+    },
+    loadImg(imgPath) {
+      return images('./' + imgPath);
```
sudeshrshetty, line 212: why need to append `./`, you can make that entry in json file only for now.

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: this is just rename  from this.cards to this.credentials

cmd/wallet-web/src/pages/mixins/common/helper.js
sudeshrshetty: why it is nested array?

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,44 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
 export const getCredentialType = (types) =>
   types.filter((type) => type != 'VerifiableCredential')[0];
 
+// function to replace json path from the json objects
+export function getCredentialDisplayData(vc, manifest) {
+  // title
+  const titleFound = findByPath(vc, manifest.title.path);
+  const title = titleFound ? titleFound : manifest.title.fallback;
```
birtony, line 96:
```suggestion
  const title = populatePath(vc, manifest.title.path[0]) || manifest.title.fallback;
```

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,44 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+  // icon
+  const icon = manifest.icon;
+
+  // find properties
+  const _readProperty = (property) => {
+    const pKey = property[0];
+    const value = findByPath(vc, property[1].path) || '';
```
birtony, line 103:
```suggestion
    const value = populatePath(vc, property[1].path[0]) || '';
```

birtony (commented):
This works for me on local

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,44 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+  return {
+    title,
+    icon,
+    properties,
+  };
+}
+// function to replace json path from the json objects
+function findByPath(vc, ...path) {
+  // returns the first value from nested path.
+  return jp.query(vc, path[0][0])[0];
```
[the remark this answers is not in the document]
birtony: Since we only update one string, we should just use `value` instead of `query` - it will update that value and return it, instead of returning an array

cmd/wallet-web/src/pages/Dashboard.vue
sudeshrshetty: I meant why you need `map(_createCard)`

cmd/wallet-web/src/pages/mixins/common/helper.js
talwinder50: its not it way too simplified in the recent option

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -129,52 +150,54 @@ export default {
...
-        let flattened = {};
-        for (let credentialKey in this.cards) {
-          this.flatten(this.cards[credentialKey].content, flattened);
-          for (let credentialContent in flattened) {
-            if (
-              obj.credentialSubject.hasOwnProperty(credentialContent) &&
-              obj.schema === this.credDisplayName(this.cards[credentialKey].content)
-            ) {
-              obj.credentialSubject[credentialContent] = flattened[credentialContent];
+      Object.values(this.credentials).map((credential) => {
```
sudeshrshetty, line 165: Why `Object.values(this.credentials).map`

why can't `this.credentials.map`?

I believe this.credentials is array now

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,39 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
 export const getCredentialType = (types) =>
   types.filter((type) => type != 'VerifiableCredential')[0];
 
+// function to get the credential display data
+export function getCredentialDisplayData(vc, manifest) {
+  // title
+  const title = populatePath(vc, manifest.title.path[0]) || manifest.title.fallback;
```
sudeshrshetty, line 95: why are you reading first path always??
should fallback to next one if first one doesn't match.

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -104,8 +124,9 @@ export default {
   },
   data() {
     return {
-      cards: [],
-      cjson: credentialDisplayData,
+      credentials: [],
+      processedCredentials: [],
+      supportedCredentials: credentialDisplayData,
```
sudeshrshetty, line 129: no need to save it as instance variable, you used it once and discard it.

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -129,52 +150,54 @@ export default {
...
+      Object.values(this.credentials).map((credential) => {
+        const currentCredentialType = this.getCredentialType(credential);
+        Object.entries(this.supportedCredentials).map(
+          ([supportedCredentialType, supportedCredential]) => {
+            if (supportedCredentialType === currentCredentialType) {
+              const processedCredential = this.getCredentialDisplayData(
+                credential,
+                supportedCredential
+              );
+              this.processedCredentials.push({ ...processedCredential, category: 'manifested' });
```
sudeshrshetty, line 174: why you need category here?

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -129,52 +150,54 @@ export default {
...
-          for (let credentialContent in flattened) {
-            if (
-              obj.credentialSubject.hasOwnProperty(credentialContent) &&
-              obj.schema === this.credDisplayName(this.cards[credentialKey].content)
-            ) {
-              obj.credentialSubject[credentialContent] = flattened[credentialContent];
+      Object.values(this.credentials).map((credential) => {
+        const currentCredentialType = this.getCredentialType(credential);
+        Object.entries(this.supportedCredentials).map(
+          ([supportedCredentialType, supportedCredential]) => {
```
sudeshrshetty, line 168: where these came from ? `[supportedCredentialType, supportedCredential]`

`credentialDisplayData` right?

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -129,52 +150,54 @@ export default {
...
-          } else {
-            flattened[key] = json[key];
-          }
-        }
+    replacer: function (key, value) {
+      if (typeof value === 'function') {
+        return value.toString();
       }
+      return value;
+    },
```
sudeshrshetty, line 198: where is this user?

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -129,52 +150,54 @@ export default {
...
+              const processedCredential = this.getCredentialDisplayData(
+                credential,
+                supportedCredential
+              );
+              this.processedCredentials.push({ ...processedCredential, category: 'manifested' });
             }
           }
+        );
+        if (!this.supportedCredentials.hasOwnProperty(currentCredentialType)) {
+          this.processedCredentials.push({ ...credential, category: 'generic' });
```
sudeshrshetty, line 179: what is category generic?

sudeshrshetty (changes requested):

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: you separate the credential we support and cred we dont , handle generic credential scenarios

cmd/wallet-web/src/pages/mixins/common/helper.js
birtony: There is only one path in each property

cmd/wallet-web/src/pages/Dashboard.vue
birtony: It does not work that way in Vue. Yet, I agree we don't need to put it in data, since it's static. We should refer to https://stackoverflow.com/questions/45565349/how-to-access-external-json-file-objects-in-vue-js-app, probably @talwinder50

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: yes

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: this was supposed to be deleted, so removed

cmd/wallet-web/src/pages/Dashboard.vue
birtony: Yes, we should just do the following:
```suggestion
      this.credentials.map((credential) => {
```

cmd/wallet-web/src/pages/Dashboard.vue
birtony: This is for credentials which we don't have specific UI for, but rather just render all properties

cmd/wallet-web/src/pages/Dashboard.vue
birtony: `replacer` is actually not used anywhere, we should get rid of it @talwinder50 @sudeshrshetty

cmd/wallet-web/src/pages/Dashboard.vue
sudeshrshetty: I suggested to use Default CredentialDIsplay data config json property as a fallback for credential type not covered.

That way UI logic will smaller

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: I have a todo for this, can do in a separate pr

cmd/wallet-web/src/pages/mixins/common/helper.js
sudeshrshetty: No, there will be multiple, her sample has just one.

birtony (approved):

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -129,52 +127,41 @@ export default {
...
-        if (json.hasOwnProperty(key)) {
-          if (json[key] instanceof Object && json[key] != '') {
-            this.flatten(json[key], flattened, key);
-          } else {
-            flattened[key] = json[key];
-          }
-        }
+    getManifest: function (credential) {
+      const currentCredentialType = this.getCredentialType(credential);
+      for (const supportedCredentialType of Object.keys(credentialDisplayData)) {
```
sudeshrshetty, line 160: Why cant we use `supportedCredentialType[currentCredentialType] || supportedCredentialType.fallback`

birtony (approved):

cmd/wallet-web/src/pages/mixins/common/helper.js
birtony: Created #1013 to address

cmd/wallet-web/src/pages/Dashboard.vue
```diff
@@ -104,8 +102,8 @@ export default {
   },
   data() {
     return {
-      cards: [],
-      cjson: credentialDisplayData,
+      credentials: [],
```
sudeshrshetty, line 105: you don't need credentials as instance variable here, not even for credential detail screen.

pass processed credential to credential detail screen not whole credential (for credential detail PR).

cmd/wallet-web/src/pages/Dashboard.vue
sudeshrshetty: just curious: why we need this for-loop here to find matching manifest?

talwinder50: you are right i tested without loop and made some change to || field and it works without loop

cmd/wallet-web/src/pages/Dashboard.vue
talwinder50: removed, make sense

cmd/wallet-web/src/pages/mixins/common/helper.js
```diff
@@ -89,6 +89,47 @@ export const isVPType = (type) => toLower(type) == 'verifiablepresentation';
...
+  };
+}
+// Populates path property in the JSON object
+function populatePath(vc, paths) {
+  for (const path of paths) {
+    try {
+      const resolvedQuery = jp.value(vc, path);
+      if (resolvedQuery) return resolvedQuery;
+    } catch (error) {
+      // TODO: write this error into logger once implemented (it would mean we received corrupt value from config file)
```
sudeshrshetty, line 127: put some console.error() for now

sudeshrshetty (approved):

PR #1292, opened 2021-12-07
fix: Adding id as prop to default and vault flyout menus

### LINKED ISSUE #1288: Check if we can set`id`s to `default` and `vault` flyout menus, since we add it as a prop
Should we also set `id`s to `default` and `vault` flyout menus, since we add it as a prop?

_Originally posted by @birtony in https://github.com/trustbloc/wallet/pull/1285#discussion_r763380043_

talwinder50:
closes #1288


talwinder50:
by mistake

PR #1474, opened 2022-02-04
feat: add didcomm V2 WACI share support

Baha-sk:
Signed-off-by: Baha Shaaban <baha.shaaban@securekey.com>

PR #1482, opened 2022-02-09
wip: didcomm v2 waci issue/share demos

Moopli:
Note:
- This PR is built on top of baha's changes in #1474
- This PR sets the wallet to use local KMS, since webKMS is currently being debugged.


Moopli:
closed this and opened #1489 instead

Issue #1527, opened 2022-02-28
Endpoint to handle OIDC Auth request


rolsonquadras:
- Add a new route
- New vue page (for now, copy WACI share page)
  - Extract PEx from claims query param 
  - Run a search through existing VCs in Wallet (use existing API)
  - Show preview/consent screen (use existing API)
  - construct response and redirect

Part of https://github.com/trustbloc/wallet/issues/1524

PR #1542, opened 2022-03-08
WIP: OIDC Share

victro:


rolsonquadras (changes requested):
We are missing the main point of this PR https://github.com/trustbloc/wallet/issues/1527. The issue is to add oidc-share flow by getting the PresentationExchange from query param and processing it. 

This PR adds DIDComm for OIDC and along with DIDComm Issuance as well.

Also, please update the commit title to be reflect specific changed than generic title ("oidc progress").

Issue #1563, opened 2022-03-14
[oidc4vp] generate vp_token


rolsonquadras:
https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0.html#name-vp_token

PR #1570, opened 2022-03-15
chore: Populating details in OIDC Share and passing "state" and token params

### LINKED ISSUE #1560: [oidc4vp] generate id_token
https://openid.net/specs/openid-connect-self-issued-v2-1_0.html#name-self-issued-id-token

### LINKED ISSUE #1562: Change OIDC Share routes to /oidc/share oidc/issue
I think eventually we will have to create the `oidc` path with a number of children routes, same as for `waci` just above

_Originally posted by @birtony in https://github.com/trustbloc/wallet/pull/1553#discussion_r824813969_

victro:
- Changing routes to /oidc/share
- Populating data without hardcoding
- Adding state, vp_token and id_token params to response
Closes #1562 
Closes #1560 

https://user-images.githubusercontent.com/2052394/159281611-4eb00cfa-b334-46ad-9b9c-168cc568cd08.mov

Part of #1563

### DIFF
--- a/cmd/wallet-web/src/pages/OIDCShare.vue
+++ b/cmd/wallet-web/src/pages/OIDCShare.vue
@@ -167,8 +167,9 @@
 import { mapGetters } from 'vuex';
 import { useI18n } from 'vue-i18n';
 import { getCredentialType, getCredentialDisplayData, getCredentialIcon } from '@/mixins';
+import { decode, encode } from 'js-base64';
 import { CredentialManager } from '@trustbloc-cicd/wallet-sdk';
-import { OIDCStore, OIDCMutations } from '@/layouts/OIDC.vue';
+import { OIDCMutations } from '@/layouts/OIDC.vue';
 import { OIDCShareLayoutMutations } from '@/layouts/OIDCShareLayout.vue';
 import StyledButton from '@/components/StyledButton/StyledButton.vue';
 import CredentialBanner from '@/components/WACI/CredentialBanner.vue';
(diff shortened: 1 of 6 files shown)


rolsonquadras:
@victro - DCO check is failing as commit sign-off is missing. Can you update the commit and push ?

cmd/wallet-web/src/pages/OIDCShare.vue
```diff
@@ -300,8 +299,50 @@ export default {
...
+          },
+          {
+            label: 'Resident Since',
+            value: '2015-01-01',
+            type: 'date',
+            format: 'yyyy-mm-dd',
+          },
+        ],
+      };
     },
```
rolsonquadras, line 346: Why is this hardcoded ? Can't we get details from the API ? Manifest is already stored as part of issuance/save flow. cc: @birtony

victro: Not hardcoding anymore with the new commit

cmd/wallet-web/src/pages/OIDCShare.vue
```diff
@@ -309,6 +346,23 @@ export default {
...
+        sub: this.$route.query.client_id,
+        aud: this.$route.query.client_id,
+        iat: Date.now(),
+        exp: Date.now(),
+      });
+      const signature = JSON.stringify({});
+      const encodedHeader = encode(header).slice(0, -1);
+      const encodedPayload = encode(payload).slice(0, -1);
+      const encodedSignature = encode(signature).slice(0, -1);
+      this.idToken = `${encodedHeader}.${encodedPayload}.${encodedSignature}`;
```
rolsonquadras, line 364: On a side note- probably we should check for jwt libraries which we can use rather than us constructing ?

victro: I tried but didn't find, maybe we should still looking for one though

sudeshrshetty (approved):

PR #1593, opened 2022-03-21
chore: Removing share success page OIDC

victro:


rolsonquadras (approved):

birtony:
Just wondering why are we removing this?

victro:
> Just wondering why are we removing this?

@birtony  We are removing this because the OIDC flow doesn't require a success page within the js client the success page will be just shown in the verifier because there would be the place the actual sharing will happen

birtony (approved):

PR #1631, opened 2022-04-02
chore: Adding OIDC Share RP registration Non-Preregistered

victro:
* Refactor ID Token generation
* Add endpoint in adapter to process registration metadata
* Adapted the flow in OIDCShare vue to match the specification


cmd/wallet-web/src/pages/OIDCShare.vue
```diff
@@ -155,6 +155,7 @@ import WACIError from '@/components/WACI/WACIError.vue';
 import WACILoading from '@/components/WACI/WACILoading.vue';
 import CredentialDetailsTable from '@/components/WACI/CredentialDetailsTable.vue';
 import OIDCShareOverview from '@/pages/OIDCShareOverview.vue';
+import JSONWebToken from '../classes/JSONWebToken';
```
birtony, line 158: We have alias defined in our `vue.config.js`, so you can use the `@` identifier:
```suggestion
import JSONWebToken from '@/classes/JSONWebToken';
```

cmd/wallet-web/src/pages/OIDCShare.vue
```diff
@@ -269,7 +270,11 @@ export default {
         console.error('get credentials failed,', e);
         this.loading = false;
       }
-      this.generateIdToken();
+      if (this.verifyRPRegistration()) {
+        this.generateIdToken();
+      } else {
+        this.errors.push('RP not registered and Id token cannot be generated');
```
birtony, line 276: Looks like we are missing a verb, maybe something like the following?
```suggestion
        this.errors.push('RP is not registered and Id token cannot be generated');
```

cmd/wallet-web/src/pages/OIDCShare.vue
```diff
@@ -327,6 +325,31 @@ export default {
       );
       this.vpToken = encodeURI(JSON.stringify(presentation));
     },
+    async verifyRPRegistration() {
+      const registrationUri = this.$route.query.registration_uri;
+      const registration = this.$route.query.registration;
```
birtony, line 330: Since `$route.query` is accessible in the url, shouldn't we sanitize the values that we read from the it? This applies to many other places in our code...
cc @sudeshrshetty @rolsonquadras

cmd/wallet-web/src/pages/OIDCShare.vue
```diff
@@ -327,6 +325,31 @@ export default {
...
+      // Just one should be given registration or registration URI
+      if (registrationUri || registration) {
+        // Non pre registered RP
+        if (registrationUri) {
+          const response = await fetch(registrationUri);
+          if (response.status === 200) {
+            // TODO: Check metadata to confirm registration
```
birtony, line 337: Could we open an issue for this `TODO` and reference it here?

cmd/wallet-web/src/pages/OIDCShare.vue
```diff
@@ -327,6 +325,31 @@ export default {
...
+            return true;
+          }
+        } else {
+          // TODO: Check metadata to confirm registration
```
birtony, line 341: Could we open an issue for this `TODO` and reference it here?

cmd/wallet-web/src/pages/OIDCShare.vue
```diff
@@ -327,6 +325,31 @@ export default {
...
+          return true;
+        }
+      } else {
+        // Pre registered RP (client_id should match given)
+        const clientIdFromRP = this.$route.query.client_id;
+        // this should be gotten from endpoint in wallet
+        const clientId = 'client_id_from_siop';
```
birtony, line 348: Are we addressing this in a future PR? Do we have an issue filed for this task?

Issue #1662, opened 2022-05-06
Rename Vue components to always be multi-word


birtony:
Our Vue components/pages naming conventions may be improved to [prevent conflicts with existing and future HTML elements](https://vuejs.org/style-guide/rules-essential.html#use-multi-word-component-names). I'm thinking we could look into renaming our components in the following folders:
- [ ] `cmd/wallet-web/src/components` - here we could add the `The` prefix to all components (`Signout` -> `TheSignout`)
- [ ] `cmd/wallet-web/src/pages` - here we could add the `Page` postfix to all components (`Vaults` -> `VaultsPage`)
- [ ] `cmd/wallet-web/src/layouts` - here we could add the `Layout` postfix to all components (`Get` -> `GetLayout`)
- [ ] `cmd/wallet-web/src/router` - here we could add the `The` prefix (`Root` -> `TheRoot`)

Suggestions on best naming conventions are welcome :slightly_smiling_face:

rolsonquadras:
@birtony @talwinder50 

> * [ ]  `cmd/wallet-web/src/components` - here we could add the `The` prefix to all components (`Signout` -> `TheSignout`)
> *

Is `The` prefix mandatory ? Or can we use any other prefix ?

birtony:
We can use any prefix @rolsonquadras

PR #1668, opened 2022-05-11
refactor: Rename Vue components to have multi-worded names

HeidiHan0000:
Closes https://github.com/trustbloc/wallet/issues/1662 

Changes:
- `cmd/wallet-web/src/components` - ~~Added `The` prefix to all component names to match Vite-powered template naming convention~~ Added the `Component` postfix to all component names

- `cmd/wallet-web/src/pages` - Added the `Page` postfix to all component names
- `cmd/wallet-web/src/layouts` - Added the `Layout` postfix to all component names
- `cmd/wallet-web/src/routes` - Added the `The` prefix to component name

Signed-off-by: @HeidiHan0000  daeun.han@avast.com


cmd/wallet-web/src/components/Vaults/TheRenameVaultModal.vue
```diff
@@ -60,13 +60,13 @@ import { computed, ref, watch } from 'vue';
...
 import { CollectionManager } from '@trustbloc/wallet-sdk';
 import { useI18n } from 'vue-i18n';
-import { vaultsMutations } from '@/pages/Vaults.vue';
-import Modal from '@/components/Modal/Modal.vue';
-import InputField from '@/components/InputField/InputField';
-import StyledButton from '@/components/StyledButton/StyledButton';
+import { vaultsMutations } from '@/pages/VaultsPage.vue';
+import Modal from '@/components/Modal/TheModal.vue';
+import InputField from '@/components/InputField/TheInputField';
+import StyledButton from '@/components/StyledButton/TheStyledButton';
```
birtony, line 66: You'd need to update the import objects as well:
```suggestion
import TheModal from '@/components/Modal/TheModal.vue';
import TheInputField from '@/components/InputField/TheInputField';
import TheStyledButton from '@/components/StyledButton/TheStyledButton';
```

birtony (commented):
This is a good progress. There are still a couple of things we need to update:
1. When you import updated vue files, you need to update the imported object name as well. And then you'd also need to update the instances of that object within the `<template>`.
2. The components inside `src/components/icons` probably don't need the `The` prefix as all of them should already have the `Icon` prefix

cmd/wallet-web/src/App.vue
```diff
@@ -9,7 +9,7 @@ import { computed, onBeforeMount, onMounted, onUnmounted, ref } from 'vue';
 import { useStore } from 'vuex';
 import getStartingLocale from '@/mixins/i18n/getStartingLocale.js';
 import { updateI18nLocale } from '@/plugins/i18n';
-import IconSpinner from '@/components/icons/IconSpinner.vue';
```
birtony, line 12: Maybe we should follow the same pattern and call our icons `spinnerIcon.vue`, `smthElseIcon.vue`, etc.

HeidiHan0000: Okay, so icon can be sort of a special component. Or could `smthIconComponent` be better for consistency..?

birtony: Hmm `-IconComponent` sounds too bulky to me...the actual meaning is that it's an icon, even though it might have some component-like features.

HeidiHan0000: I see. I've changed it to just `SpinnerIcon` and `ErrorIcon` for now, as you suggested, I think that makes sense as well

cmd/wallet-web/src/App.vue
```diff
@@ -54,7 +54,7 @@ onUnmounted(() => {
 <template>
   <div class="w-screen h-screen font-sans bg-neutrals-mischka">
     <div v-if="!loaded" class="flex relative top-1/3 flex-col justify-start items-center">
-      <IconSpinner />
+      <spinner-icon />
```
birtony, line 57: Can we use camel case for custom-defined components?
```suggestion
      <SpinnerIcon />
```

HeidiHan0000: Sure, I can change that, and that would go for all the components? 
Is there a reason why CamelCase is preferred here? I think before my changes, most were in kebab-case..

HeidiHan0000: From discussing with Anton, using PascalCase will make it more clear that it is a custom defined component, and also using kebab-case might make the template tags longer (ex `credential-details-banner-component`), which I agree with. @rolsonquadras @talwinder50 Any thoughts or can I go ahead with the PascalCase?

talwinder50: we can go head with PascalCase @HeidiHan0000

talwinder50 (dismissed):

cmd/wallet-web/src/App.vue
HeidiHan0000: Changing to PascalCase...

cmd/wallet-web/src/pages/OIDCSharePage.vue
```diff
@@ -10,19 +10,19 @@
     class="flex flex-col flex-grow justify-center items-center w-full h-full"
   >
     <!-- Loading State -->
-    <WACI-loading v-if="loading" />
+    <WACILoadingComponent v-if="loading" />
 
     <!-- Sharing State -->
-    <WACI-loading
```
HeidiHan0000, line 16: Should this use WACI-sharing? (.../components/WACI/WACISharing.vue)

talwinder50 (approved):

sudeshrshetty:
@HeidiHan0000 please squash all your commits, please amend your commit when you push a new patch.

rolsonquadras (approved):

PR #1671, opened 2022-05-18
fix: apply dynamic focus styles to credential previews

### LINKED ISSUE #1597: [DWC] Apply dynamic focus styles to credential previews
Need to apply on focus styles with `css` instead of `tailwindcss`, since it does not support dynamic styles for arbitrary values

HeidiHan0000:
Closes #1597 : Changed `CredentialPreviewComponent` to use `css` instead of `tailwindcss` for focus style 

Before (light blue focus ring on credentials that are not white)         |  After
:-------------------------:|:-------------------------:
Screen Shot 2022-05-18 at 3 27 37 PM  |  Screen Shot 2022-05-18 at 3 10 24 PM
Screen Shot 2022-05-18 at 3 27 56 PM |  Screen Shot 2022-05-18 at 3 10 16 PM
Screen Shot 2022-05-18 at 3 27 46 PM | Screen Shot 2022-05-18 at 3 09 55 PM

### DIFF
--- a/cmd/wallet-web/src/components/CredentialPreview/CredentialPreviewComponent.vue
+++ b/cmd/wallet-web/src/components/CredentialPreview/CredentialPreviewComponent.vue
@@ -6,6 +6,7 @@
 
 <template>
   <router-link
+    :id="id"
     class="
       group
       inline-flex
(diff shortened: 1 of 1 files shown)


birtony (commented):
This looks good! The only couple of things I noticed is that you need to rebase on top of main to include your most recent fix for component import and that it might be possible to pick a cleaner name for the `setColor` computed property. It is actually not a method, but a property logically, so it should be a noun, not a verb. Maybe something like `credentialStyles`?

birtony (approved):
👍

birtony:
Let's squash the commits before merging

talwinder50 (approved):

PR #1683, opened 2022-05-25
fix: [oidc4vci] web wallet view - fix OIDC router

### LINKED ISSUE #1619: [oidc4vci] web wallet view - fix OIDC router
Currently OIDC landing `/oidc` checks for query parameters to decide oidc share vs save flow.

Expected: 

- 'oidc/share' path should initiate oidc share flow, where wallet is OIDC identity provider.
- 'oidc/initiate' path should initiate oidc save flow, where wallet is OIDC client
- 'oidc/save' should be callback for authorize request flow
- Parent component of router (eg: OIDC.vue) should only have common layout related features

talwinder50:
closes #1619

https://user-images.githubusercontent.com/7862595/170584829-c90c8a83-9a27-4b6b-b1b0-21453270a236.mp4

### DIFF
--- a/cmd/wallet-web/src/layouts/OIDCInitiateLayout.vue
+++ b/cmd/wallet-web/src/layouts/OIDCInitiateLayout.vue
@@ -9,7 +9,33 @@
 </template>
 
 <script>
+import uuid from 'uuid/v4';
+import { readOpenIDConfiguration, sendCredentialAuthorizeRequest } from '@/mixins';
+import Cookies from 'js-cookie';
+
 export default {
   name: 'OIDCInitiateLayout',
+  created: async function () {
+    {
+      const opState = this.$route.query.op_state || uuid();
+      const { issuer, credential_type, manifest_id } = this.$route.query;
+      console.log('what is issuer', issuer, this.$route.query);
+      const configuration = await readOpenIDConfiguration(issuer);
+      Cookies.set(
+        opState,
+        JSON.stringify({
+          issuer,
+          credentialTypes: Array.isArray(credential_type) ? credential_type : [credential_type],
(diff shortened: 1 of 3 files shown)


cmd/wallet-web/src/router/routes.js
```diff
@@ -77,27 +77,22 @@ export default [
...
+        path: 'oidc/share',
+        name: 'oidcShare',
+        component: load('layouts/OIDCShareLayout'),
+        meta: { requiresAuth: true, signin: true, disableCHAPI: true, isNavbarHidden: true },
+      },
+      {
+        path: 'oidc/save',
+        name: 'oidcSave',
+        component: load('layouts/OIDCSaveLayout'),
+        meta: { requiresAuth: true, signin: true, disableCHAPI: true, isNavbarHidden: true },
```
birtony, line 95: This does not look right. Since `oidc/` is the base for all three of them, it would make more sense to me to have one parent `oidc` route with 3 children. Similar to how `DashboardLayout` has a number of children

talwinder50: updated

HeidiHan0000 (approved):

cmd/wallet-web/src/router/routes.js
sudeshrshetty: @birtony I remember last time you suggested to have them separated since they behave completely different when accessed, and there is no default view for `/oidc`. Can we discuss this?

sudeshrshetty (changes requested):
should decide on approach of using different routers vs same with parent-child

sudeshrshetty (approved):

PR #1711, opened 2022-06-07
feat: Initiate Gnap flow for wallet integration

### LINKED ISSUE #1678: Initiate GNAP flow - Wallet integration
We need to maintain support for the existing authentication flow while also introducing the new GNAP flow. Thus, we should:
- [ ] Add a separate route (`/gnap`) to [the Vue router](https://github.com/trustbloc/wallet/blob/main/cmd/wallet-web/src/router/routes.js) which would participate in the new authentication flow
- [ ] Add [navigation handlers in Vue route](https://github.com/trustbloc/wallet/blob/main/cmd/wallet-web/src/router/index.js#L16)r to handle necessary redirects for the new authentication flow
- [ ] Implement key pair generation logic to encrypt GNAP requests according to [the specs](https://www.ietf.org/archive/id/draft-ietf-gnap-core-protocol-09.html)

talwinder50:
closes #1678

### DIFF
--- a/cmd/wallet-web/src/mixins/gnap/gnap.js
+++ b/cmd/wallet-web/src/mixins/gnap/gnap.js
@@ -1,0 +1,32 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+import { GNAPClient } from '@trustbloc/wallet-sdk';
+
+// TODO Issue-364 Add gnap create key pair logic to agent sdk library
+export async function createKeyPair() {
+  const keyPair = await window.crypto.subtle.generateKey(
+    {
+      name: 'ECDSA',
+      namedCurve: 'P-256',
+    },
+    true,
+    ['sign', 'verify']
+  );
+
+  return keyPair;
+}
+
+export async function gnapRequestAccess(signer, gnapAuthServerURL) {
+  // TODO Issue-1709 pass the valid request to agent sdk request access
+  let req;
+  const gnapClient = new GNAPClient({
+    signer: signer,
+    gnapAuthServerURL: gnapAuthServerURL,
+  });
+  const { resp } = await gnapClient.requestAccess(req);
+  return resp;
+}
--- a/cmd/wallet-web/src/router/index.js
+++ b/cmd/wallet-web/src/router/index.js
@@ -4,8 +4,10 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
+import { computed } from 'vue';
 import { createRouter, createWebHistory } from 'vue-router';
 import store from '@/store';
+import { createKeyPair, gnapRequestAccess } from '@/mixins';
 import routes from './routes';
 
 const router = createRouter({
@@ -13,8 +15,21 @@ const router = createRouter({
   routes,
 });
 
-router.beforeEach((to, from, next) => {
+router.beforeEach(async (to, from, next) => {
   store.dispatch('agent/flushStore');
+  if (to.path === '/gnap') {
+    const gnapKeyPair = await createKeyPair();
+    store.dispatch('updateGnapKeyPair', gnapKeyPair);
+
+    const signer = { SignatureVal: gnapKeyPair };
+    const gnapAuthServerURL = computed(() => store.getters['hubAuthURL']);
+
+    const resp = await gnapRequestAccess(signer, gnapAuthServerURL);
+    // TODO Issue-1699 Save properties from resp to Vuex/local storage and call GNAPClient.continue() in separate func
+  }
+  if (to.path === 'gnap/redirect') {
+    // TODO Issue-1701 HTTP Response GNAP grant response with interact-redirect (to display list of OIDC providers)
+  }
   const locale = store.getters.getLocale;
   if (to.matched.some((record) => record.meta.requiresAuth)) {
     if (store.getters.getCurrentUser) {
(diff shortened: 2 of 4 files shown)


Moopli (approved):
can't comment on vue-router specifics, but gnap startup lgtm :+1:

cmd/wallet-web/src/mixins/gnap/gnap.js
```diff
@@ -0,0 +1,31 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+import { GNAPClient } from '@trustbloc-cicd/wallet-sdk';
+
+export async function createKeyPair() {
+  let keyPair = await window.crypto.subtle.generateKey(
```
birtony, line 10: Can we make it `const`?

cmd/wallet-web/src/router/index.js
```diff
@@ -6,15 +6,29 @@ SPDX-License-Identifier: Apache-2.0
 
 import { createRouter, createWebHistory } from 'vue-router';
 import store from '@/store';
+import { createKeyPair, gnapRequestAccess } from '@/mixins';
 import routes from './routes';
+import { computed } from 'vue/dist/vue.esm-bundler';
```
birtony, line 11:
```suggestion
import { computed } from 'vue';
import { createRouter, createWebHistory } from 'vue-router';
import store from '@/store';
import { createKeyPair, gnapRequestAccess } from '@/mixins';
import routes from './routes';
```

cmd/wallet-web/src/router/index.js
```diff
@@ -6,15 +6,29 @@ SPDX-License-Identifier: Apache-2.0
...
+router.beforeEach(async (to, from, next) => {
   store.dispatch('agent/flushStore');
+  if (to.path === '/gnap') {
+    const gnapKeypair = await createKeyPair();
+    store.dispatch('updateGnapKeyPair', gnapKeypair);
+
+    const signer = { SignatureVal: gnapKeypair };
+    const gnapAuthServerURL = computed(() => store.getters['hubAuthURL']);
+
+    const resp = await gnapRequestAccess(signer, gnapAuthServerURL);
```
birtony, line 27: So here we are going to save properties from `resp` to Vuex/local storage and call `GNAPClient.continue()`, right? Can we add a TODO with a corresponding issue in this case?

cmd/wallet-web/src/router/index.js
```diff
@@ -6,15 +6,29 @@ SPDX-License-Identifier: Apache-2.0
...
   history: createWebHistory(__webpack_public_path__),
   routes,
 });
 
-router.beforeEach((to, from, next) => {
+router.beforeEach(async (to, from, next) => {
   store.dispatch('agent/flushStore');
+  if (to.path === '/gnap') {
+    const gnapKeypair = await createKeyPair();
+    store.dispatch('updateGnapKeyPair', gnapKeypair);
```
birtony, line 22: To be consistent, can we maintain the case logic?
```suggestion
    const gnapKeyPair = await createKeyPair();
    store.dispatch('updateGnapKeyPair', gnapKeyPair);
```

cmd/wallet-web/src/router/index.js
```diff
@@ -6,15 +6,29 @@ SPDX-License-Identifier: Apache-2.0
...
+
+    const signer = { SignatureVal: gnapKeypair };
```
birtony, line 24: And here:
```suggestion
    const signer = { SignatureVal: gnapKeyPair };
```

cmd/wallet-web/src/mixins/gnap/gnap.js
```diff
@@ -0,0 +1,32 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+import { GNAPClient } from '@trustbloc-cicd/wallet-sdk';
+
+export async function createKeyPair() {
```
sudeshrshetty, line 9: key type (name, curve) to be argument & do you have long term plan to move these to wallet JS SDK?

cmd/wallet-web/src/router/index.js
```diff
@@ -4,17 +4,31 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
...
 const router = createRouter({
   history: createWebHistory(__webpack_public_path__),
   routes,
 });
 
-router.beforeEach((to, from, next) => {
+router.beforeEach(async (to, from, next) => {
   store.dispatch('agent/flushStore');
+  if (to.path === '/gnap') {
+    const gnapKeyPair = await createKeyPair();
```
sudeshrshetty, line 21: we creating key pair everytime someone access "*/gnap" in URL? this solution doesn't look scalable.
For example: you refresh this page and it creates key pair again?

sudeshrshetty (changes requested):

cmd/wallet-web/src/mixins/gnap/gnap.js
talwinder50: https://github.com/trustbloc/agent-sdk/issues/364. will move to agent sdk via this pr

cmd/wallet-web/src/router/index.js
talwinder50: as we discussed offline we are going to finalize the creation of key pair on monday with the team

talwinder50: Few lines below there is already TODO related to this added  @1699 it can be part of clientt persistance change which is already part of epic https://github.com/trustbloc/wallet/blob/53e281fab828cbd04a758d244970d0bbfe9b4b1c/cmd/wallet-web/src/router/index.js#L28

sudeshrshetty (approved):

PR #1730, opened 2022-06-17
fix: CHAPI Multiple Credential Share failure

### LINKED ISSUE #1729: CHAPI Multiple Credential Share failure
Wallet fails with following error when trying to share multiple credential through CHAPI 

> runtime-core.esm-bundler.js:226 Error: agent wasm: code: 12021, message: failed to read credential manifest: invalid credential manifest: the ID default_tbwallet_output appears in multiple output descriptors
>     at agent.js:1:1560

The latest wallet code deployed in dev fails with [this](https://github.com/trustbloc/sandbox/blob/main/docs/demo/new-bank-account-usecase.md) usecase.

Possible issue:
The output descriptors [here](https://github.com/trustbloc/wallet/blob/4b2d9826aaf36a57c1174f025a499a2559bd8ec9/cmd/wallet-web/src/mixins/common/helper.js#L335) contain duplicate ids when multiple credentials don't have descriptors defined. We need to see if removing duplicated would fix the issue.

HeidiHan0000:
Closes #1729 

Was able to reproduce locally by:
- Removing all credential output descriptors but "`VerifiableCredential`" (which is the default if the output descriptor cannot be found) in `credential-output-descriptors.json`
- In `readManifests` in `options.js`, comment [this line](https://github.com/trustbloc/wallet/blob/4b2d9826aaf36a57c1174f025a499a2559bd8ec9/cmd/wallet-web/src/store/modules/options.js#L396) so the local output descriptor file gets used.
- Save a PR card and University credential to wallet
- Use MultipleQuery 1 to query for a PR card and University credential - the error is seen here

Fix: To use a Set to add only unique `output_descriptors` to `credentialManifest`

### DIFF
--- a/cmd/wallet-web/src/mixins/common/helper.js
+++ b/cmd/wallet-web/src/mixins/common/helper.js
@@ -301,6 +301,7 @@ export function prepareCredentialManifest(presentation, manifestDictionary, issu
     version: '0.1.0',
     output_descriptors: [],
   };
+  let uniqueOutputDescriptors = new Set();
 
   let fulfillment = {
     id: uuidv4(),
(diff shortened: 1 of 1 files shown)


rolsonquadras (approved):
Nice work 👍

birtony (approved):

PR #1736, opened 2022-06-21
chore: added check for null before adding to localStorage

### LINKED ISSUE #1331: Unexpected behaviour with the store
We currently use `null` as a default value for most of the properties in our `Vuex` state. We also save those default values to the `localStorage`. However, according to [this post,](https://stackoverflow.com/questions/15193461/how-to-set-localstorage-item-back-to-null), it is not possible to save a `null` value to the `localStorage`, so all of the `null` values are saved as strings. Afterwards, when we `loadUser` (in `user.js`) we get values from the `localStorage` and assign them to `Vuex` properties. As a result, our `Vuex` properties have `null`s stored as strings instead of actual `null`.

In order to avoid this unintended conversion, we can check if the item exists in `localStorage` and return actual `null` if it does not:
```ls
localStorage.getItem('someValue') === 'null' ? null : localStorage.getItem('someValue');
```

cc @sudeshrshetty

HeidiHan0000:
Closes #1331 

Added check for null before adding to `localStorage`. Then when `localStorage.getItem` is called, it will return `null` (not string 'null') if the key does not exist in `localStorage`. 

### DIFF
--- a/cmd/wallet-web/src/store/modules/user.js
+++ b/cmd/wallet-web/src/store/modules/user.js
@@ -25,23 +25,33 @@ export default {
   mutations: {
     setUser(state, val) {
       state.username = val;
-      localStorage.setItem('user', val);
+      if (val !== null) {
+        localStorage.setItem('user', val);
+      }
     },
     setProfile(state, val) {
       state.profile = val;
-      localStorage.setItem('profile', JSON.stringify(val));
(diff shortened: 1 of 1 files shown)


birtony (approved):
👍🏼

rolsonquadras (approved):

PR #1784, opened 2022-07-12
fix: dev mode for wallet

aholovko:
Signed-off-by: Andrii Holovko <andriy.holovko@gmail.com>


birtony:
@aholovko it'd probably make sense to have two targets in the root Makefile: 
- One for running linter in fix mode (so that devs can lint their code locally before pushing)
- One for running linter in check mode in CI (so that it reports any errors related to code not linted without attempting to lint ititself)

PR #1786, opened 2022-07-12
fix: Update provider file to use only gnap provider

talwinder50:


HeidiHan0000 (approved):

rolsonquadras (changes requested):
We need to remove oidc setup config as well - https://github.com/trustbloc/wallet/blob/main/test/fixtures/wallet-web/hydra-config/demo_hydra_configure.sh

rolsonquadras (approved):

PR #1847, opened 2022-08-11
feat: added flag to enable CHAPI in wallet

### LINKED ISSUE #1837: Add feature flag to enable CHAPI
- [ ] Add a flag to Wallet Setup https://github.com/trustbloc/wallet/blob/main/images/wallet-web/Dockerfile#L42
- [ ] Install/Register CHAPI only if flag is true

birtony:
Fixes #1837

https://user-images.githubusercontent.com/33902374/184799236-6778871c-662b-41ba-97de-c37cc44c3c05.mov

### DIFF
--- a/test/ui-automation/package-lock.json
+++ b/test/ui-automation/package-lock.json
@@ -19,7 +19,7 @@
         "@babel/preset-env": "^7.18.10",
         "@babel/register": "^7.18.9",
         "@wdio/local-runner": "^7.20.9",
-        "@wdio/mocha-framework": "^7.20.3",
+        "@wdio/mocha-framework": "^7.23.0",
         "@wdio/spec-reporter": "^7.21.0"
       }
     },
@@ -2411,15 +2411,15 @@
       }
     },
     "node_modules/@wdio/mocha-framework": {
-      "version": "7.20.3",
-      "resolved": "https://registry.npmjs.org/@wdio/mocha-framework/-/mocha-framework-7.20.3.tgz",
-      "integrity": "sha512-ZOY+MvOap8yrckRg86CIPu+H1ibSJp6nrKhkQqVsVUO/JLvUlVYUHJm8pVngrfbVtLljoldXuNdnnxm7oUNR9g==",
+      "version": "7.23.0",
+      "resolved": "https://registry.npmjs.org/@wdio/mocha-framework/-/mocha-framework-7.23.0.tgz",
+      "integrity": "sha512-/qMc9FEl5B1CWfBBd/vJR6gbZpBSk71hR+29ME00RflE3RhlDHlJkXnGQi1cDd+jjp8LQyxG9+yZmVdUDBQaQA==",
       "dev": true,
       "dependencies": {
         "@types/mocha": "^9.0.0",
         "@wdio/logger": "7.19.0",
-        "@wdio/types": "7.20.3",
-        "@wdio/utils": "7.20.3",
+        "@wdio/types": "7.23.0",
+        "@wdio/utils": "7.23.0",
         "expect-webdriverio": "^3.0.0",
         "mocha": "^10.0.0"
       },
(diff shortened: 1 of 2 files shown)


cmd/wallet-web/src/main.js
```diff
@@ -16,11 +15,29 @@ import ToastNotificationComponent from '@/components/ToastNotification/ToastNoti
...
-app.use(i18n);
-app.provide('polyfill', polyfill);
-app.provide('webCredentialHandler', webCredentialHandler);
-app.component('ToastNotification', ToastNotificationComponent);
+(async () => {
+  if (process.env.NODE_ENV === 'production') {
+    // call service to get the agent opts
+    const CHAPIEnabled = await axios
+      .get(`${window.location.origin}/walletconfig/agent`)
+      .then((resp) => resp.data.enableCHAPI)
```
rolsonquadras, line 23: We already make a call to get wallet-config options before we load the first page. This adds a duplicate network call to fetch same data and adds to overall performance. Can we update code to have only one call to this API ?

cmd/wallet-web/src/router/index.js
```diff
@@ -71,7 +71,30 @@ router.beforeEach(async (to, from) => {
...
+      store.dispatch('updateUserLoaded', true);
+      return {
+        name: 'NotFound',
+        params: {
+          title:
+            'The page you requested requires CHAPI, but your Wallet configuration does not support it at the moment.',
+          message: 'Please, contact your administrator.',
+        },
+      };
+    }
```
rolsonquadras, line 94: Can we simply this block to something like below?

>     if (store.getters.getEnableCHAPI === undefined && store.dispatch('loadUser')) {
>        await store.dispatch('initOpts');
>      } else if (!store.getters.getEnableCHAPI) {
>        store.dispatch('updateUserLoaded', true);
>        return {
>          name: 'NotFound',
>          params: {
>            title:
>              'The page you requested requires CHAPI, but your Wallet configuration does not support it at the moment.',
>            message: 'Please, contact your administrator.',
>          },
>        };
>      }

cmd/wallet-web/src/store/modules/options.js
```diff
@@ -138,50 +139,52 @@ export default {
         readCredentialManifests = readManifests();
       }
 
-      const optValue = (opt) => (opt in agentOpts ? agentOpts[opt] : defaultAgentStartupOpts[opt]);
+      const chooseOptValue = (opt) =>
```
rolsonquadras, line 142: Any specific reason to rename this ?

cmd/wallet-web/src/router/index.js
```diff
@@ -71,7 +71,30 @@ router.beforeEach(async (to, from) => {
...
-  } else if (to.matched.some((record) => record.meta.requiresAuth)) {
+  }
+
+  if (to.matched.some((record) => record.meta.requiresCHAPI)) {
+    // Get enableCHAPI from wallet config
+    if (store.getters.getEnableCHAPI === undefined) {
+      if (store.dispatch('loadUser') && store.getters.getEnableCHAPI === undefined) {
+        await store.dispatch('initOpts');
+      }
+    }
```
rolsonquadras, line 82: Can we simplify this block to something like below?

>     if (store.getters.getEnableCHAPI === undefined && store.dispatch('loadUser')) {
>        await store.dispatch('initOpts');
>      }

birtony: The logic behind is to firstly try to get the value from Vuex, secondly fetch it from the local storage, and lastly make an axios call. As soon as the value is not `undefined`, we won't make any further steps. There are edge cases, when the value is not available in vuex or local storage, so we use api call as a last resort.
@rolsonquadras @sudeshrshetty

cmd/wallet-web/src/router/index.js
```diff
@@ -71,7 +71,30 @@ router.beforeEach(async (to, from) => {
...
+
+    if (!store.getters.getEnableCHAPI) {
+      store.dispatch('updateUserLoaded', true);
+      return {
+        name: 'NotFound',
+        params: {
+          title:
+            'The page you requested requires CHAPI, but your Wallet configuration does not support it at the moment.',
+          message: 'Please, contact your administrator.',
```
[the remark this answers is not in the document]
birtony: There are no French string provided for the `Not Found` page as of now since we don't have the designs. As discussed, this is a temporary page to be replaced/updated when the designs are provided

cmd/wallet-web/src/store/modules/options.js
birtony: Since it is a function, it sound cleaner to me to use the verb in its name. Essentially, this function chooses which option to use @rolsonquadras @sudeshrshetty

cmd/wallet-web/src/main.js
birtony: Good point, but at in `main.js` we don't yet have the Vue app initialized, so no store, router, etc. - we can only save the fetches values in local storage. However, I'm thinking we might be able to move the dynamic import/provide logic altogether to a later stage somewhere in `App.vue` and utilize the store from there. I am looking into this possibility

cmd/wallet-web/src/store/modules/options.js
Moopli: fine by me :+1:

PR #1848, opened 2022-08-11
feat: added check for empty optional properties for resolveManifest

### LINKED ISSUE #1832: Wallet should handle cases where optional properties are missing from resolved credential manifest
Ensure wallet can handle cases where optional fields returned from resolveManifest do not have valid values. The credential manifest specs specify certain properties of output descriptors as mandatory and others as optional, and out of the properties returned from resolveManifest, there are fields which may not have valid values. As this is valid as per the docs, wallet should be able to handle cases where those fields are empty, using default values for those properties.

HeidiHan0000:
Closes #1832 
Added check for empty optional fields in the manifest before it gets resolved by `resolveManifest`. These are fields that are required by the wallet to display the credential properly but are not required in the credential manifest as per the [specs](https://identity.foundation/wallet-rendering/v0.0.1/#entity-styles) (ex. credential name, background colour, etc...).  

### DIFF
--- a/cmd/wallet-web/src/mixins/common/helper.js
+++ b/cmd/wallet-web/src/mixins/common/helper.js
@@ -357,3 +357,83 @@ export function prepareCredentialManifest(presentation, manifestDictionary, issu
   credentialManifest.output_descriptors = Array.from(uniqueOutputDescriptors);
   return credentialManifest;
 }
+
+/**
+ * Check credential manifest for missing optional properties before resolving to ensure the
+ * resolved manifest has all fields required for wallet to display the credential properly
+ *
+ *  @param {Object} credentialManager - Credential Manager to resolve manifest.
+ *  @param {String} auth - authorization token for wallet operations.
+ *  @param {Object} options - options to resolve credential from wallet.
+ *  @param {String} options.credentialID - (optional) ID of the credential to be resolved from wallet content store.
+ *  @param {String} options.credential - (optional) raw credential data model to be resolved.
+ *  @param {String} options.fulfillment - (optional) credential fulfillment using which given raw credential or credential ID to be resolved.
+ *  @param {String} options.manifestID - (optional) ID of the manifest from wallet content store.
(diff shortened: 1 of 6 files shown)


rolsonquadras (approved):

cmd/wallet-web/src/mixins/common/helper.js
```diff
@@ -357,3 +357,70 @@ export function prepareCredentialManifest(presentation, manifestDictionary, issu
...
+    });
+  }
+  manifest.output_descriptors[0] = outputDescriptor;
+
+  return manifest;
+}
+
+function getDefaultOutputDescriptor(credentialType) {
+  const defaultManifests = require('@/config/credential-output-descriptors.json');
+  if (!defaultManifests.hasOwnProperty(credentialType)) {
```
birtony, line 411:
```suggestion
  if (!Object.prototype.hasOwnProperty(defaultManifests, credentialType)) {
```

cmd/wallet-web/src/mixins/common/helper.js
```diff
@@ -357,3 +357,70 @@ export function prepareCredentialManifest(presentation, manifestDictionary, issu
...
+    console.error('default credential manifest for credential type does not exist');
+  }
+  const defaultManifest = defaultManifests[credentialType];
+  const outputDescriptor = defaultManifest[Object.keys(defaultManifest)[0]].output_descriptors[0];
```
birtony, line 415: Do we know what `Object.keys(defaultManifest)[0]` or is it dynamic? Just thinking if we could just do `defaultManifest.someKnownPropertyName` instead

HeidiHan0000: These are different urls for examples of the different vcs, for example for PR card : 
```
 {
  "PermanentResidentCard": {
    "https://w3id.org/citizenship/v1": {
        "output_descriptors": [
        { 
             ... different fields...
         }
      }
  }
```
So they are different for each credential

birtony (approved):
LGTM

PR #1862, opened 2022-08-23
feat: added flag to enable CHAPI in wallet

### LINKED ISSUE #1837: Add feature flag to enable CHAPI

birtony:
Fixes #1837

Note that currently we are still making 2 calls to `walletconfig` on sign up/in, because existing logic fetches both wallet config and gnap bootstrap option from the same function `initOpts`. We should probably split it up when we refactor the store, and address this at that point.

### CHAPI Disabled Error Page
image

### VC Wallet Mode Sign In/Out
https://user-images.githubusercontent.com/33902374/186503794-794d3d9f-b9ed-4483-a8b2-0523b9875f2a.mov

### DIDComm Disabled Error Page (already existed)
image


cmd/wallet-web/src/App.vue
```diff
@@ -5,27 +5,77 @@
...
+    console.log('username', username.value);
+    chapi
+      .install(username.value)
+      .then(() => activateCHAPI())
+      .catch((e) => {
+        console.error(e);
+      });
+  }
+});
+
```
sudeshrshetty, line 78: not sure why we need CHAPI install in App.vue when we can do it after login (or init after new page/tab) triggered by vuex state.

cmd/wallet-web/src/components/Signout/SignoutComponent.vue
```diff
@@ -29,17 +26,21 @@ const signoutUser = () => store.dispatch('logout');
...
   if (isCHAPI.value) {
+    const polyfill = await import('credential-handler-polyfill');
+    const webCredentialHandler = await import('web-credential-handler');
+    const agentOpts = computed(() => store.getters['getAgentOpts']);
+
+    const chapi = ref(
+      new CHAPIHandler(polyfill, webCredentialHandler, agentOpts.value.credentialMediatorURL)
+    );
+
     actions.push(chapi.value.uninstall());
```
sudeshrshetty, line 37: can we do it in better way?
triggering an vuex action and that action finds CHAPI if it's there and uninstall it.

I would suggest all CHAPI handling (install, register, uninstall) isolated and called only by concerned actions.

birtony: This is the logic we used up until now, I just added a conditional statement here to support the flag feature. I totally agree with your suggestion to clean this up, but maybe we can address it after we overview wallet data/vuex model and architecture?

cmd/wallet-web/src/App.vue
```diff
@@ -5,27 +5,77 @@
...
+// const activateCHAPI = () => store.dispatch('activateCHAPI');
+
+// if intended target doesn't require CHAPI or if it is disabled
+// const enableCHAPI = computed(
+//   () =>
+//     CHAPIEnabled.value &&
+//     !route.params?.disableCHAPI &&
+//     (route.name === 'chapi-worker' || (!breakpoints.xs && !breakpoints.sm))
+// );
+// console.log('enableCHAPI', enableCHAPI.value);
```
rolsonquadras, line 42: Can we remove commented code ?

Moopli (approved):

sudeshrshetty (approved):

PR #1866, opened 2022-08-29
feat: pick mysql image based on architecture type

### LINKED ISSUE #1865: Use corresponding mysql image based on architecture type

birtony:
Fixes #1865

### DIFF
--- a/scripts/wallet_web_start.sh
+++ b/scripts/wallet_web_start.sh
@@ -8,5 +8,11 @@ set -e
 
 echo "Running $0"
 
+# Use corresponding image based on architecture type (arm64/amd64)
+arch=$(uname -p)
+if  [[ $arch == arm64 ]]; then
+    export MYSQL_IMAGE=arm64v8/mysql
+fi
+
 cd test/fixtures/wallet-web
 (source .env && docker-compose -f docker-compose-demo.yml -f docker-compose-server.yml -f docker-compose-web.yml down && docker-compose -f docker-compose-demo.yml -f docker-compose-server.yml -f docker-compose-web.yml up --force-recreate)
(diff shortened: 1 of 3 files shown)


talwinder50 (dismissed):

scripts/wallet_web_start.sh
```diff
@@ -8,5 +8,11 @@ set -e
 
 echo "Running $0"
 
+# Use corresponding image based on architecture type (AMD/ARM)
```
DRK3, line 11: Should be called `amd64` or `x86` instead of `AMD` technically

DRK3: or `x86-64`

scripts/wallet_web_start.sh
```diff
@@ -8,5 +8,11 @@ set -e
...
+arch=$(uname -p)
+if  [[ $arch == arm* ]]; then
```
DRK3, line 13: I think we should check for an exact architecture match since there are technically different `arm` architectures, and the MySQL build below is for `arm64`

scripts/wallet_web_start.sh
DRK3: (I would recommend `x86` or `x86-64` as they are the most commonly understood names)

DRK3: See https://en.wikipedia.org/wiki/X86-64 if you're curious where the name `amd64` came from... it's because AMD originally created the 64-bit extension to the x86 instruction set

DRK3: Looks like MySQL uses the term `amd64` on their [docker page](https://hub.docker.com/_/mysql), so perhaps that's the best one to use here to ensure clarity.

DRK3 (approved):

rolsonquadras (approved):

PR #1871, opened 2022-08-31
test: updated configs to run both vc wallet and regular wallet instances

### LINKED ISSUE #1838: Configure Credential Exchange mechanism - Integration tests
- [ ] setup 2 instances of Wallet 
  1. With no DIDCOMM and CHAPI flags
  2. With DIDCOMM=true and CHAPI=true
- [ ] update automation tests to pass 2 wallet urls 
- [ ] run tests based on configured wallet url

birtony:
Fixes #1838

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,7 +100,42 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-  UIAutomation:
+  VCWallet-e2e:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Run Wallet Web UI automation
+        run: |
+          echo '127.0.0.1 testnet.orb.local' | sudo tee -a /etc/hosts
+          echo '127.0.0.1 vcwallet.trustbloc.local' | sudo tee -a /etc/hosts
+          echo '127.0.0.1 mediator.trustbloc.local' | sudo tee -a /etc/hosts
+          echo '127.0.0.1 uni-resolver-web.trustbloc.local' | sudo tee -a /etc/hosts
+          echo '127.0.0.1 auth.trustbloc.local' | sudo tee -a /etc/hosts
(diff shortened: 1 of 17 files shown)


test/ui-automation/test/helpers/wallet.js
```diff
@@ -18,8 +18,12 @@ let signedUpUserEmail;
 exports.signUp = async ({ email }) => {
   // login and consent
   await _getSignUp(email);
-  // register chapi
-  await allow();
+
+  if (browser.config.walletName === "wallet.trustbloc.local:8091") {
```
rolsonquadras, line 22: Rather than this, can we pass a bool to this function to skip this part ? This way we can keep helpers cleaner. The bool param can from calling function. Also, lets not hardcode the value in the code.

.github/workflows/build.yml
```diff
@@ -100,7 +100,42 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Run Wallet Web UI automation
+        run: |
+          echo '127.0.0.1 testnet.orb.local' | sudo tee -a /etc/hosts
+          echo '127.0.0.1 vcwallet.trustbloc.local' | sudo tee -a /etc/hosts
```
rolsonquadras, line 115: We need to update doc as well https://github.com/trustbloc/wallet/blob/main/docs/build.md#connections

scripts/run_vcwallet_ui_automation.sh
```diff
@@ -54,7 +54,7 @@ healthCheck() {
 }
 
 # healthcheck
-healthCheck wallet-web https://wallet.trustbloc.local:8091/healthcheck 200
+healthCheck vcwallet https://vcwallet.trustbloc.local:8071/healthcheck 200
```
talwinder50, line 57: if we are replacing wallet-web what will be the another instance which will run with chapi and didcomm enabled ?

birtony: https://github.com/trustbloc/wallet/pull/1871/files#diff-bf6f61dd6a378d5262de4cb22c38fcd438d891d9ecadb9526f625e1461c39d33R9-R61

rolsonquadras (approved):

DRK3 (approved):

